PR #18, opened 2018-08-01
Add sabakan installer script

ueokande:
Embed installer script `/usr/local/sabakan/install` into the container.  This script installs client apps (`sabactl` and `sabakan-cryptsetup`) and bash completion to the host file system.  For example, install these files by rkt as the following:

```console
$ sudo rkt run \
    --volume host-usr-local-bin,kind=host,source=/usr/local/bin,readOnly=false \
    --mount volume=host-usr-local-bin,target=/host/usr/local/bin \
    --volume host-etc-bash-completion-d,kind=host,source=/etc/bash_completion.d,readOnly=false \
    --mount volume=host-etc-bash-completion-d,target=/host/etc/bash_completion.d \
    quay.io/cybozu/sabakan:0 \
    --user 0 \
    --exec /usr/local/sabakan/install-tools
```


sabakan/.dockerignore
```diff
@@ -1 +1 @@
-*
+!installer
```
dulltz: `!install` ?

ueokande:
fixed

PR #87, opened 2018-11-27
[sabakan] Update to 0.25

dulltz:


ymmt2005:
sabakan 0.25 should be skipped because it was released without incrementing version number in `version.go`.

PR #141, opened 2019-02-25
[circleci] skip pushing branch tags for pre-releases

ymmt2005:
Pre-release images are not official ones therefore should
not update branch tags.  Pre-releases can be identified by
detecting "-" in the tag string.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,10 @@ jobs:
                 TAG=$(cat $d/TAG)
                 docker tag quay.io/cybozu/${d}:latest quay.io/cybozu/${d}:$TAG
                 docker push quay.io/cybozu/${d}:$TAG
+                if ! echo $TAG | grep -q -e - ; then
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -49,6 +49,10 @@ jobs:
                 TAG=$(cat $d/TAG)
                 docker tag quay.io/cybozu/${d}:latest quay.io/cybozu/${d}:$TAG
                 docker push quay.io/cybozu/${d}:$TAG
+                if ! echo $TAG | grep -q -e - ; then
```
mitsutaka: 条件逆でしょうか

```
$ echo "1.2rc.1" | grep -q -e -; echo $?
1
$ echo "1.2-rc.1" | grep -q -e -; echo $?
0
```

ymmt2005: あ、うっかり。
```suggestion
                if echo $TAG | grep -q -e - ; then
```

PR #205, opened 2019-07-12
[teleport] add teleport container

binoue:
This pull request made the following changes.
- add files to create teleport containers 
- modify circle ci config to create teleport container image in CI

Notes:
- teleport does not provide the way to create teleport container image in its github explicitly. Therefore, this Docker file is referencing to this Dockerfile( https://github.com/gravitational/teleport/blob/v4.0.2/build.assets/charts/Dockerfile ).
- In the circle CI, we newly ignore docker lint error DL3005 which means  "Do not use apt-get upgrade or dist-upgrade" because the referenced Dockerfile is using `apt-get upgrade` inside.


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
             fi
       - run:
           name: Lint Dockerfile
-          command: docker run --rm -i hadolint/hadolint /bin/hadolint - --ignore DL3008 --ignore DL3016 < << parameters.container-image >>/Dockerfile
+          command: docker run --rm -i hadolint/hadolint /bin/hadolint - --ignore DL3008 --ignore DL3016 --ignore DL3005 < << parameters.container-image >>/Dockerfile
```
zoetrope: DL3005は無視するのではなく、Dockerfileのほうを直してください

teleport/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN git checkout v${TELEPORT_VERSION} && \
+    GO111MODULE=off make full
+
+# Stage2: final image
+FROM quay.io/cybozu/ubuntu:18.04
+
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get install --no-install-recommends -y \
+    dumb-init \
+    ca-certificates \
```
zoetrope: ca-certificatesはすでに入ってる

PR #226, opened 2019-09-06
[kind-node] add kind-node container

kfyharukz:


kind-node/src/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/cybozu/neco-containers/kind-node/src
+
+go 1.12
+
+require (
+	github.com/containernetworking/cni v0.7.1
+	github.com/containernetworking/plugins v0.8.2
```
zoetrope: Should we use v0.7.5？

PR #244, opened 2019-10-23
Update golang 1.13

llamerada-jp:
Update golang 1.13 at containers.


ymmt2005:
Publishing rebuilt container images is not necessary.

PR #268, opened 2019-12-19
admission: Add mutating/validating webhooks for Contour HTTPProxy.

morimoto-cybozu:
This PR adds mutating/validating webhooks for Contour `HTTPProxy`.

The annotations of `HTTPProxy` objects will be mutated/validated.
The mutating webhook enforces a default annotation of `kubernetes.io/ingress.class: <configured value>` if an `HTTPProxy` object is not annotated with the keys of `kubernetes.io/ingress.class` or `projectcontour.io/ingress.class`.
The validating webhook denies an `HTTPProxy` object if it is not annotated with the keys.

The default value is given through the `--httpproxy-default-class` option.


admission/hooks/mutate_httpproxy.go
```diff
@@ -0,0 +1,59 @@
...
+	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	annotationKubernetesIngressClass = "kubernetes.io/ingress.class"
+	annotationContourIngressClass    = "projectcontour.io/ingress.class"
+	annotationIngressClassDefault    = "forest"
```
ymmt2005: It would be better if this is not a constant, but a variable that
is given by a command-line flag because:

- `forest` is too specific.
- We may change our minds.

admission/hooks/validate_httpproxy.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// Note: An empty class name is not safe if the implementation of Ingress controller considers it as "not specified" and starts serving.
+	if hp.Annotations[annotationKubernetesIngressClass] != "" {
+		return admission.Allowed("ok")
+	}
+	if hp.Annotations[annotationContourIngressClass] != "" {
+		return admission.Allowed("ok")
+	}
+
+	return admission.Denied(fmt.Sprintf("either %s or %s annotation should be set in %s/%s", annotationKubernetesIngressClass, annotationContourIngressClass, hp.Namespace, hp.Name))
```
ymmt2005: `hp.Name` or `hp.Namespace` is often empty.  Use `req.Name` or `req.Namespace` instead.

PR #275, opened 2020-01-16
coredns: Change base image from scratch to ubuntu

tapih:
https://github.com/cybozu-go/neco/issues/661
To use the `sleep` command in k8s `preStop` hook, the base images in some containers should be changed to ubuntu.


contour/Dockerfile
```diff
@@ -15,7 +15,7 @@ WORKDIR /work/contour/
 RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOLDFLAGS="-w -s" go build -o /go/bin/contour -v github.com/projectcontour/contour/cmd/contour
```
ymmt2005: Since CoreDNS runs on Ubuntu, `CGO_ENABLED=0` for static binary is no longer needed.

coredns/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN curl -sSLf https://github.com/coredns/coredns/archive/v${COREDNS_VERSION}.ta
 WORKDIR /go/src/github.com/coredns/coredns/
 RUN make
 
-FROM scratch
+FROM quay.io/cybozu/ubuntu:18.04
 
 COPY --from=build /etc/ssl/certs /etc/ssl/certs
```
ymmt2005: The new base image already has ca-certificates.

PR #278, opened 2020-01-23
admission: add validator for ArgoCD Application to enforce spec.project.

morimoto-cybozu:
Add the new validating Webhook for ArgoCD `Application` CR.

`Application.spec.project` is declared to control the authority what resources tenants can use.
This field is a relationship with `AppProject` CR.

However, if we allow tenants to create `Application`, `spec.project` can be mutated by tenants.
It will cause the breaking authority separation.
We want to allow tenants to create `Application`,
but we do not want to allow tenants to set their project only to `spec.project`.

To achieve this, we implement the new admission Webhook.

If this PR have merged, `neco-admission` will become to need the configuration file at the path of `--config-path` .

#### Example of configuration file.
```yaml
ArgoCDApplicationValidator:
  rules:
    - repository: https://github.com/cybozu-private/maneki-apps.git
      projects:
        - maneki
```


admission/docs/configuration.md
```diff
@@ -0,0 +1,29 @@
...
+
+| Name       | Type       | Description                                                               |
+| ---------- | ---------- | ------------------------------------------------------------------------- |
+| repository | string     | A URL of the repository                                                   |
+| projects   | \[\]string | A list of `spec.project`s allowed for the applications in the repository. |
+
+```yaml
+ArgoCDApplicationValidator:
+  rules:
+    - repository: https://github.com/cybozu-private/maneki-apps.git
```
ymmt2005: What happens if the Application has a repository URL w/o `.git` suffix?

dulltz: It will not matched the repoURL, but this behavior is not prefer

dulltz: We decided to ignore `.git` prefix.

admission/hooks/validate_application.go
```diff
@@ -0,0 +1,73 @@
...
+	app.SetGroupVersionKind(schema.GroupVersionKind{Group: "argoproj.io", Kind: "Application", Version: "v1alpha1"})
+	err := v.decoder.Decode(req, app)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	repoURL, found, err := unstructured.NestedString(app.UnstructuredContent(), "spec", "source", "repoURL")
+	if !found {
+		return admission.Errored(http.StatusBadRequest, errors.New("spec.source.repoURL not found"))
+	}
+	if err != nil {
```
ymmt2005: Shouldn't `err` need to check first before `found`?

And non-nil error means the value is not an expected type, here `string`.

admission/hooks/validate_application.go
```diff
@@ -0,0 +1,73 @@
...
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("unable to get spec.rource.repoURL; %w", err))
+	}
+	project, found, err := unstructured.NestedString(app.UnstructuredContent(), "spec", "project")
+	if !found {
+		return admission.Errored(http.StatusBadRequest, errors.New("spec.project not found"))
+	}
+	if err != nil {
```
ymmt2005: ditto

admission/hooks/validate_application.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	repoURL, found, err := unstructured.NestedString(app.UnstructuredContent(), "spec", "source", "repoURL")
+	if !found {
+		return admission.Errored(http.StatusBadRequest, errors.New("spec.source.repoURL not found"))
+	}
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("unable to get spec.rource.repoURL; %w", err))
+	}
+	project, found, err := unstructured.NestedString(app.UnstructuredContent(), "spec", "project")
+	if !found {
```
ymmt2005: Isn't `spec.project` optional?

dulltz: `spec.project` is not optional. It is required.

dulltz: So it might not be necessary this `found` check,
but I'll remain the check because it is possible to change the CRD.

admission/docs/configuration.md
```diff
@@ -0,0 +1,37 @@
...
+
+Each rule represents the restriction on the applications in a certain repository.  
+If neco-admission has no rule for a given App's repoURL, neco-admission denies the API request.
+
+| Name       | Type       | Description                                                               |
+| ---------- | ---------- | ------------------------------------------------------------------------- |
+| repository | string     | A URL of the repository to be matched with `applications.spec.source.repoURL`.                                                   |
+| projects   | \[\]string | A list of `applications.spec.project`s allowed for the applications in the repository. |
+
+##### `.git` suffix in `repository`
```
ymmt2005: Too deep.
```suggestion
### `.git` suffix in `repository`
```

PR #284, opened 2020-01-31
[rook] Change the base image.

kfyharukz:


rook/Dockerfile
```diff
@@ -22,7 +22,7 @@ WORKDIR ${ROOK_DIR}/images/ceph
 RUN make generate-csv-ceph-templates
 
 # Stage2: setup runtime container
-FROM ceph/ceph-amd64:v14.2.6-20200115
```
ymmt2005: This can be given by a command-line argument as shown in this article:
https://www.jeffgeerling.com/blog/2017/use-arg-dockerfile-dynamic-image-specification

It'd be better to parameterize the base image, especially, the image tag `:14.2` and supply
the value in CircleCI config.

.circleci/config.yml
```diff
@@ -400,6 +405,8 @@ workflows:
       - build:
           name: build-rook
           container-image: rook
+          requires:
+            - build-ceph-container
```
ymmt2005: This seems not necessary, because uploading to `quay.io` happens only in the `master` branch.

rook/Dockerfile
kfyharukz: I think the info about what version to use should be separated from CircleCI config.
So, the base image name and its version should be written in each Dockerfile, shouldn't it?
(I know some exceptions already exist, e.g. envoy and ceph...

ymmt2005: I thought so because ceph version is specified in circleci config.

ymmt2005: That said, this is minor.  I'll mark this as resolved.

.circleci/config.yml
kfyharukz: I understood. Applied.

PR #285, opened 2020-02-03
metrics-server: fix Dockerfile to use git clone

llamerada-jp:
https://github.com/cybozu-go/neco/issues/484


metrics-server/Dockerfile
```diff
@@ -3,10 +3,15 @@ FROM quay.io/cybozu/golang:1.13-bionic AS build
 
 ARG METRICS_SERVER_VERSION=0.3.6
 ARG CGO_ENABLED=0
```
ymmt2005: ARG is not ENV.

PR #292, opened 2020-02-17
admission: add mutation webhook to mount emptyDir on /tmp

dulltz:
*Pod mutator* mutates Pods to create emptyDir and mount it on `/tmp` if any volume does not exist which is mounted under `/tmp`.

The purpose of this webhook is to avoid sudden death due to writing under read-only filesystem.


admission/hooks/mutate_pod.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+	return false
+}
+
+func (m *podMutator) hashString(name string) string {
+	h := sha1.New()
+	h.Write([]byte(name))
+	bs := h.Sum(nil)
+	return fmt.Sprintf("%x", bs)
```
ymmt2005: This can be written simply as:
```go
        hex.EncodeToString(sha1.Sum([]byte(name)))
```

admission/hooks/mutate_pod.go
```diff
@@ -0,0 +1,101 @@
...
+	"strconv"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:verbs=create;update,path=/mutate-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,versions=v1,name=mpod.kb.io
```
ymmt2005: I don't think `update` verb is needed.

admission/hooks/mutate_pod.go
masa213f: Thanks. I fixed it.
And the hash string was too long, so I cut it to 16 characters.

admission/hooks/mutate_pod.go
masa213f: Thanks. I fixed it.

admission/hooks/mutate_pod.go
ymmt2005: Would you please describe why it is limited to 16 characters?
And if there are any clear reasons, please add comments on the code.

masa213f: The limitation(16 characters) has no meaning.

I checked the hash suffix generated by Kubernetes, it is 5 characters for Pod.
So, I will change the length to 5 characters just like a Pod.

ymmt2005: If there is no reason, I'd like to keep the hash string as is (40 bytes?).

masa213f: OK. I will remove the limitation. The generated volume name is as follows. (`tmp-`+55 charactoers)
```
tmp-7562756e747530da39a3ee5e6b4b0d3255bfef95601890afd80709
```

admission/hooks/mutate_pod.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, mount := range co.VolumeMounts {
+		if mount.MountPath == "/tmp" || strings.HasPrefix(mount.MountPath, "/tmp/") {
+			return true
+		}
+	}
+	return false
+}
+
+func (m *podMutator) hashString(name string) string {
+	return hex.EncodeToString(sha1.New().Sum([]byte(name)))
```
ymmt2005:
```suggestion
	sum := sha1.Sum([]byte(name))
	return hex.EncodeToString(sum[:])
```

admission/hooks/mutate_pod.go
ymmt2005: That's strange...  SHA1 is 20 bytes, which will be converted to a 40 byte hex string.

admission/hooks/mutate_pod.go
masa213f: I understood.
My fix was wrong. The hash value could not be calculated.

admission/hooks/mutate_pod.go
ymmt2005: Got it.  Your usage of `sha1.New().Sum(...)` was wrong.  I fixed it.

PR #298, opened 2020-03-04
envoy: update to v1.12.3

masa213f:
- Update envoy to v1.12.3 as the vulnerability response.
- To make the build faster, the CircleCI resource class will be `2xlarge+`.


masa213f:
> @masa213f
> Please increase `jobs` to match the new number of CPU cores.
> https://github.com/cybozu/neco-containers/blob/master/.circleci/config.yml#L216

Thanks. I will increase the `jobs` to `20`. Because `2xlarge+` has 20 vCPUs.
https://circleci.com/docs/2.0/configuration-reference/#docker-executor

PR #314, opened 2020-03-27
cke-tools: add updateblock117 for upgrading publish path of Kubelet

dulltz:
Add `updatedevice117` command to cke-tools.

Spec document is here: [README.md#updateblock117](https://github.com/cybozu/neco-containers/blob/updateblock/cke-tools/src/README.md#updateblock117)


cke-tools/src/README.md
```diff
@@ -18,3 +19,27 @@ CKE tools is licensed under MIT license.
 
 [godoc]: https://godoc.org/github.com/cybozu/neco-containers/cke-tools/src
 [CKE]: https://github.com/cybozu-go/cke
+
+`updateblock117`
+----------------
```
ymmt2005: Move this section before the list of markdown references.
References should always placed at the end of the document.

cke-tools/src/README.md
```diff
@@ -18,3 +19,27 @@ CKE tools is licensed under MIT license.
...
+
+`updateblock117` has two subcommands.
+
+### `updateblock117 need-update <block-pv-name>`
+
+Check that we should modify the path of the target device file or not.
+
+`block-pv-name` is the name of the PersistentVolume object.
+
+- Returns stdout `{"result": "yes"}` if the device needs to tweak.
```
ymmt2005: > - Returns stdout `{"result": "yes"}` if the device needs to tweak.
> - Returns stdout `{"result": "no"}` if the device needs to tweak.

Probably one is wrong. Note that `if the device needs to tweak.` appears twice.
Also, the text sounds unnatural.  `when the device file needs to be relocated` is better.

cke-tools/src/cmd/updateblock117/cmd/needupdate.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/cybozu-go/cke-tools/pkg/updateblock117"
+	"github.com/spf13/cobra"
+)
+
+var needUpdateCmd = &cobra.Command{
+	Use:   "need-update <block-pv-name>",
+	Short: "check that we should modify the path of the target device file or not",
+	Long:  "check that we should modify the path of the target device file or not",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
```
ymmt2005: `RunE` would display the usage text if not silenced.

Add
```go
                // without this, each subcommand's RunE would display usage text.
                cmd.SilenceUsage = true
```

cke-tools/src/cmd/updateblock117/cmd/operate.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/cybozu-go/cke-tools/pkg/updateblock117"
+	"github.com/spf13/cobra"
+)
+
+var operateCmd = &cobra.Command{
+	Use:   "operate <block-pv-name>",
+	Short: "move block device to new location and fix symbolic link",
+	Long:  "move block device to new location and fix symbolic link",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
```
ymmt2005: ditto

cke-tools/src/cmd/updateblock117/cmd/operate.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/cybozu-go/cke-tools/pkg/updateblock117"
+	"github.com/spf13/cobra"
+)
+
+var operateCmd = &cobra.Command{
+	Use:   "operate <block-pv-name>",
```
ymmt2005: I don't think using `<>` is appropriate for this.
```suggestion
	Use:   "operate PV",
```

cke-tools/src/cmd/updateblock117/cmd/needupdate.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/cybozu-go/cke-tools/pkg/updateblock117"
+	"github.com/spf13/cobra"
+)
+
+var needUpdateCmd = &cobra.Command{
+	Use:   "need-update <block-pv-name>",
```
ymmt2005:
```suggestion
	Use:   "need-update PV",
```

cke-tools/src/cmd/updateblock117/cmd/needupdate.go
```diff
@@ -0,0 +1,71 @@
...
+			return err
+		}
+		if existsOld {
+			res.Result = "yes"
+			out, err := json.Marshal(res)
+			if err != nil {
+				return err
+			}
+			_, err = fmt.Println(string(out))
+			return err
```
ymmt2005: The same code block appears 3 times.  Could you refactor it?

cke-tools/src/cmd/updateblock117/cmd/root.go
```diff
@@ -0,0 +1,26 @@
...
+
+// Execute executes the root command.
+func Execute() error {
+	return rootCmd.Execute()
+}
+
+func init() {
+	rootCmd.AddCommand(needUpdateCmd)
+	rootCmd.AddCommand(operateCmd)
+}
```
ymmt2005: Subcommands should be registered in each file.  Not here.

cke-tools/src/pkg/updateblock117/operate.go
```diff
@@ -0,0 +1,55 @@
...
+	targetPath := makeNewDeviceFilePath(pvName, podUID)
+	symlinkPath := filepath.Join(deviceRootDir, pvName, "dev", podUID)
+	symlinkTmpPath := symlinkPath + ".tmp"
+
+	err = os.Remove(symlinkTmpPath)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	err = os.Symlink(targetPath, symlinkTmpPath)
```
ymmt2005: Why don't you create the symlink directly as
```go
err = os.Symlink(targetPath, symlinkPath)
```

dulltz: It is because we should update symlink atomically.
os.Symlink cannot update existence symlink file.

PR #323, opened 2020-04-06
[2020-04] Update the former half of images

llamerada-jp:
This is the former half of the container image update on April 2020.
The latter half is https://github.com/cybozu/neco-containers/pull/324.

Updating ceph is out of scope of this PR.


envoy/TAG
```diff
@@ -1 +1 @@
-1.12.3.1
+1.13.1.1
```
tapih, line 1: The version of `envoy` is updated both in TAG and BRANCH, but the version in .circleci/config.yaml is not updated.

admission/TAG
```diff
@@ -1 +1 @@
-0.4.2
```
ymmt2005, line 1: Since neco-admission container is from scratch, there is no need to update TAG.

gorush/TAG
```diff
@@ -1 +1 @@
-emiksk-1.11.2.2
```
ymmt2005, line 1: gorush is not managed by us.  It is managed by Maneki.
Ignore it.

local-pv-provisioner/TAG
```diff
@@ -1 +1 @@
-0.1.1
```
ymmt2005, line 1: Just a comment; this is to update k8s client modules.
If we had updated the modules, local-pv-provisioner need not be updated because its image is scratch-based.

bmc-reverse-proxy/TAG
```diff
@@ -1 +1 @@
-0.1.2
```
ymmt2005, line 1: Just a comment; this is to update k8s client modules.
If we had updated the modules, bmc-reverse-proxy need not be updated because its image is scratch-based.

machines-endpoints/TAG
```diff
@@ -1 +1 @@
-0.4.0
```
ymmt2005, line 1: Just a comment; this is to update k8s client modules.
If we had updated the modules, machines-endpoints need not be updated because its image is scratch-based.

admission/TAG
llamerada-jp: We agree with you. However, we noticed that contour version was not updated yet. Eventually, we need to update TAG in this PR.

gorush/TAG
llamerada-jp: We reverted.

ymmt2005:
Please rebase your commits to remove commits for `gorush` completely.

PR #327, opened 2020-04-10
contour: fix contour binary path

umezawatakeshi:
According to upstream Dockerfile, contour binary is placed at `/bin/contour` and the manifest file of ingress https://github.com/cybozu-go/neco-apps/blob/1ce6103b21fad83575dffc91a1d8b0bcfa539c40/ingress/base/template/deployment.yaml#L144 execute contour as `/bin/contour`


ymmt2005:
Why is this fix needed?

PR #333, opened 2020-04-22
ceph: update ceph base image

binoue:


ceph/README.md
```diff
@@ -14,13 +14,27 @@ To use in Rook, you need to write a manifest of the custom resource CephCluster
...
 ----------------------------
 
-This image based on the Dockerfile produced by `ceph/ceph-container`.
-You can make the Dockerfile with following commands:
-```console
-git clone git@github.com:ceph/ceph-container.git
-cd ceph-container
-make FLAVORS=luminous,debian,9 stage
-# Then the Dockerfile will be located at ./staging/luminous-debian-9-x86_64/daemon-base
+This container uses a cybozu's own Ceph [cybozu/ceph][]:neco-release. It's because using dmcrypt devices that is already encrypted by users is not supported in the upstream's newest stable version (v15.2.1), but the function is  necessary for Neco.
```
ysksuzuki, line 17: > a cybozu's own Ceph
It sounds like Cybozu has ownership of Ceph. I think customized, enhanced or patched Ceph is better in this context.

ceph/README.md
```diff
@@ -14,13 +14,27 @@ To use in Rook, you need to write a manifest of the custom resource CephCluster
...
+# Please set $CEPH_VERSION, e.g. CEPH_VERSION="14.2.8.4"
+cd go/src/github.com/ceph/ceph
+git checkout master && git pull
+git remote add fork git@github.com:cybozu/ceph.git
+git fetch fork && git checkout neco-release
+git rebase fork/master
+# Please resolve conflict & dependencies carefully
+
+git push -f fork neco-release
+git tag "v$CEPH_VERSION"
```
zoetrope, line 33: You should annotate the tag.
This is because ceph uses `git describe` to extract the version.

binoue: Sorry, I forgot that

PR #336, opened 2020-04-23
envoy: add liveness probe to combine multiple liveness.

morimoto-cybozu:
This PR adds a liveness probe for Envoy.
This probe is intended to be deployed as a sidecar container in an Envoy Pod.

This probe listens on `localhost:8502/`.
It monitors the following endpoints and replies OK only if all of the endpoints are healthy.
- `localhost:8002/ready`: original readiness probe of Envoy
- `localhost:8080`: HTTP service port; the returned status code is not checked
- `localhost:8443`: HTTPS service port; only the TCP connection is checked

The endpoint addresses are configurable via command line options.


envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,168 @@
...
+
+	if conn != nil {
+		conn.Close()
+	}
+	return nil
+}
+
+func init() {
+	fs := rootCmd.Flags()
+	fs.StringVar(&config.listenAddr, "listen-addr", ":8003", "Listen address for probe")
```
ymmt2005, line 163: I think this is too close to the Envoy's readiness check port 8002.
Shouldn't we choose a more distant number?

envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,168 @@
...
+			"status": resp.StatusCode,
+		})
+		return fmt.Errorf("readiness probe returned %d", resp.StatusCode)
+	}
+
+	return nil
+}
+
+func monitorHTTP(ctx context.Context) error {
+	c := well.HTTPClient{
```
ymmt2005, line 120: Share the same `http.Client` with `monitorReady`.  It is designed so.

envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,168 @@
...
+		return
+	}
+
+	rw.WriteHeader(http.StatusOK)
+	rw.Write([]byte("ok"))
+	return
+}
+
+func monitorReady(ctx context.Context) error {
+	c := well.HTTPClient{
```
ymmt2005, line 88: Do not throw away http.Client.  The package is not designed so.
https://golang.org/pkg/net/http/#Client

> The Client's Transport typically has internal state (cached TCP connections), so **Clients should be reused** instead of created as needed. Clients are safe for concurrent use by multiple goroutines.

Also, use a private http.Transport and configure it not to keep-alive connections.
```go
    tp := &http.Transport{
        DisableKeepAlives: true,

        // rest are copied from http.DefaultTransport
        Proxy: ProxyFromEnvironment,
        DialContext: (&net.Dialer{
            Timeout:   30 * time.Second,
            KeepAlive: 30 * time.Second,
            DualStack: true,
        }).DialContext,
        ForceAttemptHTTP2:     true,
        MaxIdleConns:          100,
        IdleConnTimeout:       90 * time.Second,
        TLSHandshakeTimeout:   10 * time.Second,
        ExpectContinueTimeout: 1 * time.Second,
    }
```

envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,168 @@
...
+
+	env.Stop()
+	err := env.Wait()
+	if err != nil {
+		rw.WriteHeader(http.StatusBadGateway)
+		rw.Write([]byte(err.Error()))
+		return
+	}
+
+	rw.WriteHeader(http.StatusOK)
```
ymmt2005, line 82: No need for this.
`Write` w/o calling `WriteHeader` will return response 200.
In fact, both `WriteHeader` and `Write` are not necessary.

envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,168 @@
...
+	env := well.NewEnvironment(req.Context())
+	env.Go(monitorReady)
+	env.Go(monitorHTTP)
+	env.Go(monitorHTTPS)
+
+	env.Stop()
+	err := env.Wait()
+	if err != nil {
+		rw.WriteHeader(http.StatusBadGateway)
+		rw.Write([]byte(err.Error()))
```
ymmt2005, line 78: Set `Content-Type` header appropriately.

envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+func init() {
+	fs := rootCmd.Flags()
+	fs.StringVar(&config.listenAddr, "listen-addr", ":8003", "Listen address for probe")
+	fs.DurationVar(&config.timeout, "timeout", time.Second*5, "Timeout")
+	fs.StringVar(&config.readyURL, "ready-url", "http://0.0.0.0:8002/ready", "URL of Envoy readiness probe")
+	fs.StringVar(&config.httpURL, "http-url", "http://0.0.0.0:8080/", "URL for checking HTTP behavior")
+	fs.StringVar(&config.httpsAddr, "https-addr", "0.0.0.0:8443", "Address for checking HTTPS behavior")
```
ymmt2005, line 167: I think `localhost` is more suitable than `0.0.0.0` because `localhost` can work in IPv6 only environments.

envoy/Dockerfile
```diff
@@ -1,11 +1,22 @@
...
+RUN CGO_ENABLED=0 GO111MODULE=on go install -mod=vendor -ldflags="-w -s" ./pkg/livenessprobe
+
+# Stage2: setup runtime container
 FROM quay.io/cybozu/ubuntu:18.04
 
 COPY workspace/envoy /usr/local/bin/envoy
 COPY workspace/docker-entrypoint.sh /docker-entrypoint.sh
 COPY workspace/LICENSE /usr/local/share/doc/envoy/LICENSE
 COPY envoy.yaml /etc/envoy/envoy.yaml
+COPY --from=build /go/bin /
```
ymmt2005, line 19: Since envoy is installed under `/usr/local/bin`, why not does this too?

envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,188 @@
...
+	httpsAddr string
+}
+
+var rootCmd = &cobra.Command{
+	Use:   "livenessprobe",
+	Short: "liveness probe for Envoy",
+	Long:  `Liveness probe for Envoy.`,
+
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		cmd.SilenceUsage = true
```
ymmt2005, line 38: This is only needed when you use `RunE` instead of `Run`.

envoy/pkg/livenessprobe/cmd/root.go
```diff
@@ -0,0 +1,188 @@
...
+	env.Go(m.monitorReady)
+	env.Go(m.monitorHTTP)
+	env.Go(m.monitorHTTPS)
+
+	env.Stop()
+	err := env.Wait()
+	if err != nil {
+		log.Error("returning failure result", map[string]interface{}{
+			log.FnError: err,
+		})
```
ymmt2005, line 110: This duplicates the error logs.

envoy/pkg/livenessprobe/cmd/root.go
ymmt2005: Also, using `PersistentPreRun` is a little overkill because there is only one
subcommand (rootCmd).  The logic here can be moved inside `Run`.

PR #386, opened 2020-06-11
ingress-watcher: Ingress monitoring tool

kfyharukz:
This PR implements an ingress monitoring tool.
See: https://github.com/cybozu-go/neco/issues/1005


ingress-watcher/pkg/watch/watch_test.go
```diff
@@ -0,0 +1,176 @@
...
+	httpGetFailTotalName       = "ingresswatcher_http_get_fail_total"
+)
+
+func TestWatcherRun(t *testing.T) {
+	type fields struct {
+		targetURLs []string
+		interval   time.Duration
+		httpClient *http.Client
+	}
+	type args struct {
```
dulltz, line 31: `args` is not used

ingress-watcher/pkg/watch/watch_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+type RoundTripFunc func(req *http.Request) (*http.Response, error)
+
+func (f RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req)
+}
+
+func newTestClient(fn RoundTripFunc) *http.Client {
+	return &http.Client{
+		Transport: RoundTripFunc(fn),
```
dulltz, line 166: This conversion is not necessary

ingress-watcher/pkg/watch/watch_test.go
```diff
@@ -0,0 +1,176 @@
...
+				tt.fields.httpClient,
+			)
+			env := well.NewEnvironment(context.Background())
+			env.Go(func(ctx context.Context) error {
+				ctx, cancel := context.WithTimeout(ctx, timeoutDuration)
+				defer cancel()
+				return w.Run(ctx)
+			})
+			env.Stop()
+			env.Wait()
```
dulltz, line 97: Why do not you check error retuned by `env.Wait()`?

ingress-watcher/cmd/export.go
```diff
@@ -0,0 +1,80 @@
...
+		if len(exportConfig.ListenAddr) == 0 {
+			return errors.New("required flag \"listen-addr\" not set")
+		}
+		return nil
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		well.Go(watch.NewWatcher(
+			rootConfig.TargetURLs,
+			rootConfig.WatchInterval,
+			&http.Client{},
```
dulltz, line 47: Why do not you use `well.HTTPClient`?
https://godoc.org/github.com/cybozu-go/well#HTTPClient

ingress-watcher/cmd/push.go
```diff
@@ -0,0 +1,90 @@
...
+			return errors.New("required flag \"push-addr\" not set")
+		}
+
+		return nil
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		well.Go(watch.NewWatcher(
+			rootConfig.TargetURLs,
+			rootConfig.WatchInterval,
+			&http.Client{},
```
dulltz, line 48: ditto

ingress-watcher/pkg/watch/watch_test.go
dulltz: You can write as `Transport: fn,`

ingress-watcher/cmd/root.go
```diff
@@ -0,0 +1,65 @@
...
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func init() {
+	fs := rootCmd.PersistentFlags()
+	fs.StringArrayVarP(&rootConfig.TargetURLs, "target-urls", "", nil, "Target Ingress address and port.")
+	fs.DurationVarP(&rootConfig.WatchInterval, "watch-interval", "", 5*time.Second, "Watching interval.")
+	fs.StringVarP(&configFile, "config", "", "", "Configuration YAML file path.")
```
zoetrope, line 58: The usage says that the `configFile` is a global configuration.
However, the format of the configuration file is different for each subcommand.
I think this option is confusing.

ingress-watcher/cmd/push.go
```diff
@@ -0,0 +1,90 @@
...
+	Short: "Push metrics to Pushgateway",
+	Long:  `Push metrics to Pushgateway`,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if configFile != "" {
+			if err := viper.Unmarshal(&pushConfig); err != nil {
+				return err
+			}
+		}
+
+		if len(pushConfig.JobName) == 0 {
```
zoetrope, line 34: Couldn't you use MarkFlagRequired?

ingress-watcher/cmd/push.go
```diff
@@ -0,0 +1,90 @@
...
+		err := well.Wait()
+		if err != nil && !well.IsSignaled(err) {
+			log.ErrorExit(err)
+		}
+	},
+}
+
+func init() {
+	fs := pushCmd.Flags()
+	fs.StringVarP(&pushConfig.PushAddr, "push-addr", "", "", "Pushgateway addres.")
```
zoetrope, line 85: typo

ingress-watcher/cmd/push.go
```diff
@@ -0,0 +1,90 @@
...
+	Long:  `Push metrics to Pushgateway`,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if configFile != "" {
+			if err := viper.Unmarshal(&pushConfig); err != nil {
+				return err
+			}
+		}
+
+		if len(pushConfig.JobName) == 0 {
+			return errors.New("required flag \"job-name\" not set")
```
zoetrope, line 35: I prefer to use single-quotes or back-quotes rather than escaping double-quotes.

ingress-watcher/cmd/push.go
```diff
@@ -0,0 +1,90 @@
...
+		well.Go(func(ctx context.Context) error {
+			tick := time.NewTicker(pushConfig.PushInterval)
+			defer tick.Stop()
+
+			for {
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-tick.C:
+					pusher := push.New(pushConfig.PushAddr, pushConfig.JobName).Gatherer(registry)
```
zoetrope, line 59: Do you need to make this pusher every time in the loop?

ingress-watcher/pkg/watch/watch_test.go
tapih: If the error is checked here, the error caused by timeout is also captured (the timeout behavior is intended).
I think the timeout error can be separated from the other errors, but I don't do that because of simplicity.

ingress-watcher/cmd/push.go
```diff
@@ -0,0 +1,90 @@
...
+					err := pusher.Add()
+					if err != nil {
```
zoetrope, line 61: If there is an error, such as a failure to connect to pushgateway, can we notice the error?

ingress-watcher/cmd/push.go
tapih: This is because I cannot find the way to set flag as required when the value can be given from both YAML file and command line flag.

For example, when I set MarkFlagRequired and configure a value with YAML file, viper gets an error which says the value is not set even if the value is definitely set.

ingress-watcher/cmd/push.go
tapih: There is a metric that records the last pushed timestamp.
So, if the watcher continues to fail pushing the metrics, the timestamp gets old and we can notice that by alert.

ingress-watcher/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+2. Push and expose the collected metrics via [Pushgateway](https://github.com/prometheus/pushgateway).
+    ```bash
+    ingress-watcher push \
+    --target-urls https://example.com \
+    --target-urls https://example.org \
+    --push-addr localhost:9091 \
+    --watch-interval 10s \
+    --job-name job \
+    --push-interval 5s
```
zoetrope, line 30: Why can you set the `--watch-interval` and `--push-interval` to different values?
It will cause a duplicate value push.
I think pushing should be at the same time as watching.

kfyharukz: `--watch-interval` indicates a sampling rate of checking Ingress. On the other hand, `--push-interval` means updating interval of metrics' counter.

So, for example, a setting such as smaller `--watch-interval` and larger `--push-interval` make sense (the example in `README.md` is not good, so I'll rewrite it).

ingress-watcher/cmd/root.go
kfyharukz: I think so.  Therefore, I modified them to make the config file accepted only for each sub command, and settings in `root.go` were moved to each config.

PR #399, opened 2020-06-26
ingress-watcher: Add a resolve-rule option

llamerada-jp:
Signed-off-by: llamerada <Yuji Ito>


llamerada-jp:
closed because design of ingress-watcher was changed

PR #422, opened 2020-09-09
coredns: Upgrade to 1.7.0

yokaze:
Upgrade CoreDNS from v1.6.7 to v1.7.0.
Some metrics names changed, but none of them are used.


morimoto-cybozu (approved):
mobbed

PR #428, opened 2020-09-10
rook: update rook version to 1.4.2 with a patch

llamerada-jp:


kfyharukz (approved):
LGTM

PR #429, opened 2020-09-11
kubernetes: manager#syncPod should cover reconciliation

umezawatakeshi:
apply https://github.com/kubernetes/kubernetes/pull/89155


morimoto-cybozu (approved):
paired

PR #434, opened 2020-09-17
ci: check go.mod in cke-tools

yokaze:

### DIFF
--- a/cke-tools/src/Makefile
+++ b/cke-tools/src/Makefile
@@ -2,7 +2,11 @@ SUDO = sudo
 SUDO_GO = $(SUDO) $(shell which go)
 
 # Run tests
-test:
+check-generate:
+	go mod tidy
+	git diff --exit-code --name-only
+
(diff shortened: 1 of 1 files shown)


morimoto-cybozu (approved):
LGTM

cke-tools/src/Makefile
```diff
@@ -2,7 +2,11 @@ SUDO = sudo
 SUDO_GO = $(SUDO) $(shell which go)
 
 # Run tests
-test:
+check-generate:
+	go mod tidy
+	git diff --exit-code --name-only
+
+test: check-generate
```
ymmt2005, line 9: You should separately run `check-generate` from normal test,
because `check-generate` has side-effects.

@morimoto-cybozu @yokaze

morimoto-cybozu: @ymmt2005 
I missed that point.
I'll fix it.

PR #447, opened 2020-10-19
Add an option to ignore restrictpkg

zoetrope:


kfyharukz (approved):
~~LGTM~~

golang-bionic/analyzer/pkg/restrictpkg/testdata/src/ignore/ignore.go
```diff
@@ -0,0 +1,29 @@
+package a
```
kfyharukz, line 1: The package name should be `ignore`, shouldn't it?

golang/analyzer/pkg/restrictpkg/testdata/src/ignore/ignore.go
```diff
@@ -0,0 +1,29 @@
+package a
```
kfyharukz, line 1: ditto

kfyharukz (changes requested):

golang-bionic/analyzer/pkg/restrictpkg/testdata/src/ignore/ignore.go
zoetrope: I think it should be `main`. Because it has a main function.

kfyharukz (approved):
LGTM

PR #462, opened 2020-11-26
use original VictoriaMetrics operator 0.4.0

umezawatakeshi:
I find that operator is able to patch container specs that are created by operator. So our patch is not necessary.


morimoto-cybozu (approved):
LGTM

PR #463, opened 2020-11-26
Refactor makefiles and test all go sources

masa213f:
- Fix how to specify `.PHONY` target
- Unify the test job of CiercleCI
- Use `staticcheck` instead of `golint`
- Test all go sources (and fix)

### DIFF
--- a/admission/Makefile
+++ b/admission/Makefile
@@ -13,30 +13,37 @@ GOARCH := $(shell go env GOARCH)
 GOFLAGS = -mod=vendor
 export GOFLAGS
 
+.PHONY: all
 all: build
 
 # Run tests
+.PHONY: test
 test: crds
 	test -z "$$(gofmt -s -l . | grep -v '^vendor' | tee /dev/stderr)"
-	test -z "$$(golint $$(go list ./... | grep -v /vendor/) | grep -v '/mtest/.*: should not use dot imports' | tee /dev/stderr)"
+	staticcheck ./...
 	test -z "$$(nilerr $$(go list ./... | grep -v /vendor/) 2>&1 | tee /dev/stderr)"
 	test -z "$$(custom-checker -restrictpkg.packages=html/template,log ./... 2>&1 | grep -v vendor/ | tee /dev/stderr)"
 	ineffassign .
-	go test -race -v ./...
 	go vet ./...
+	go test -race -v ./...
+
 
 # Build manager binary
+.PHONY: build
 build:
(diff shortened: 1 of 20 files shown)


admission/Makefile
```diff
@@ -13,9 +13,11 @@ GOARCH := $(shell go env GOARCH)
 GOFLAGS = -mod=vendor
 export GOFLAGS
 
+.PHONY: all
 all: build
 
 # Run tests
+.PHONY: test
 test: crds
 	test -z "$$(gofmt -s -l . | grep -v '^vendor' | tee /dev/stderr)"
```
tapih, line 22: Can this block be replaced with `$(MAKE) -f $(COMMON_MAKEFILE ) test`??

local-pv-provisioner/Makefile
```diff
@@ -45,7 +40,7 @@ generate:
 check-uncommitted:
 	$(MAKE) manifests
```
tapih, line 41: Can the `manifests` and `generate` targets be written as the dependent target, like `check-uncommitted: manifests generate`??

tapih (commented):
Basically, LGTM. But might need some tweaks.

admission/Makefile
masa213f: Thank you. I removed `Makefile.common` because it would be needlessly complicated.

local-pv-provisioner/Makefile
masa213f: Yes. They can be written as the dependent target.
But I don't change it because it makes the intentions clearer, I think.

cke-tools/src/rivers/server.go
```diff
@@ -56,7 +56,8 @@ func NewServer(upstreams []string, cfg Config) *Server {
 		dialer:    dialer,
 		pool: sync.Pool{
 			New: func() interface{} {
-				return make([]byte, copyBufferSize)
+				buf := make([]byte, copyBufferSize)
+				return &buf
```
zoetrope, line 60: Why did you turn it into a pointer?

zoetrope: For performance?
https://qiita.com/peroxyacyl/items/5e02ddf4480ecd2ec7b1

zoetrope: - SA6002 – Storing non-pointer values in sync.Pool allocates memory
  https://staticcheck.io/docs/checks#SA6002

zoetrope (approved):
LGTM

PR #468, opened 2020-12-01
Downgrade gorush 1.12.0 (Build with goalng 1.15)

yokotaso:
APNs won't work well with 1.13.0.

Ref: https://github.com/appleboy/gorush/issues/536

Therefore I update only golang.


masa213f (approved):
LGTM

PR #474, opened 2020-12-10
admission: Add validate-service webhook

tapih:
This is a workaround for [CVE-2020-8554](https://github.com/kubernetes/kubernetes/issues/97076), which confirms the `externalIPs` field is empty.
Signed-off-by: H.Muraoka <hiroshi-muraoka@cybozu.co.jp>


admission/hooks/validate_service.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (v *serviceValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	svc := &corev1.Service{}
+	err := v.decoder.Decode(req, svc)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if svc.Spec.Type == corev1.ServiceTypeClusterIP && len(svc.Spec.ExternalIPs) > 0 {
```
ymmt2005, line 34: `externalIPs` can be specified for any type of Services.
https://kubernetes.io/docs/concepts/services-networking/service/#external-ips

I think Services of other types, for example type=NodePort, should also deny externalIPs.

ymmt2005 (approved):
LGTM

admission/hooks/validate_service_test.go
```diff
@@ -0,0 +1,60 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ = Describe("test for allowing rules for Service", func() {
+	It("should allow creating Service with externalIPs empty", func() {
+		svc := &corev1.Service{}
+		svc.Name = "allow1"
+		svc.Namespace = "default"
+		svc.Spec.Type = corev1.ServiceTypeClusterIP
```
champtar, line 14: You should test with other types as well, see https://github.com/kubernetes/kubernetes/issues/97076#issuecomment-740317872 (the validation is ok for me)

tapih: Thank you for telling me that.
I already understood that any other service type than ClusterIP should be validated out,
but the logic of the webhook is so simple that we just left the test as it is now.

PR #479, opened 2020-12-14
Remove tail container

masa213f:
We don't use the tail container.
It was created for MOCO's log rotation, but the filebeat container is now used.


yokaze:
@masa213f 
Looks good, but it has conflicting files.
Can you please rebase it and make a review request again?

yokaze (approved):
LGTM

PR #482, opened 2020-12-16
upgrade building go version  for testhttpd

binoue:


ymmt2005:
This does nothing.

What we should do is to build testhttpd with Go 1.15, isn't it?

testhttpd/TAG
```diff
@@ -1 +1 @@
-0.1.0
```
ymmt2005, line 1: Since this does not change anything, the tag should be `0.1.1`.

ymmt2005 (approved):
LGTM

ymmt2005:
Build failed.  Let's stop vendoring.
```
 ---> Running in c359f2e92ae5
go: inconsistent vendoring in /work/src:
	github.com/cybozu-go/log@v1.5.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/cybozu-go/well@v1.10.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
```

ymmt2005 (approved):
LGTM

PR #485, opened 2020-12-22
Update cert-manager API ver to v1

kfyharukz:
Signed-off-by: Kazuhito MATSUDA <kazuhito.matuda@gmail.com>

### DIFF
--- a/bmc-reverse-proxy/bmc-reverse-proxy.yaml
+++ b/bmc-reverse-proxy/bmc-reverse-proxy.yaml
@@ -49,7 +49,7 @@ spec:
   selector:
     app.kubernetes.io/name: bmc-reverse-proxy
 ---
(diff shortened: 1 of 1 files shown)


binoue (commented):
No need to change the apiVersion below?
- https://github.com/cybozu/neco-containers/blob/update-cert-manager-api-ver/admission/config/certmanager/certificate.yaml#L3
- https://github.com/cybozu/neco-containers/blob/update-cert-manager-api-ver/admission/config/certmanager/certificate.yaml#L11

kfyharukz:
@binoue 
Those resources of `cert-manager` were generated by `kubebuilder` and were not used in `neco-apps` acctually.  So, I think we don't need to update those resources.

binoue (approved):
LGTM

PR #495, opened 2021-01-12
rook: add a patch to resolve the issue that ob is not created from obc using ArgocD

toshipp:
Signed-off-by: Toshikuni Fukaya <toshikuni-fukaya@cybozu.co.jp>


llamerada-jp (approved):
mobbed

PR #497, opened 2021-01-15
admission: Fix condition for min order in vnetworkpolicy.

kfyharukz:
This PR fixes the condition for min order in `vnetworkpolicy` from "less than and equal" to only "less than".
It means the policy that has the order `1000` will be allowed, while the current version denies it.


ymmt2005:
Update `TAGS` too.

admission/cmd/run.go
```diff
@@ -46,7 +46,7 @@ func run(addr string, port int, conf *hooks.Config) error {
 	dec, _ := admission.NewDecoder(scheme)
 	wh := mgr.GetWebhookServer()
 	wh.Register("/mutate-pod", hooks.NewPodMutator(mgr.GetClient(), dec))
-	wh.Register("/validate-projectcalico-org-networkpolicy", hooks.NewCalicoNetworkPolicyValidator(mgr.GetClient(), dec, 1000))
```
ymmt2005, line 49: I don't like this change...
Please keep this line, and change elsewhere.

admission/hooks/run_test.go
```diff
@@ -39,7 +39,7 @@ func run(stopCh <-chan struct{}, cfg *rest.Config, opts *envtest.WebhookInstallO
 	dec, _ := admission.NewDecoder(scheme)
 	wh := mgr.GetWebhookServer()
 	wh.Register(podMutatingWebhookPath, NewPodMutator(mgr.GetClient(), dec))
-	wh.Register(calicoValidateWebhookPath, NewCalicoNetworkPolicyValidator(mgr.GetClient(), dec, 1000))
+	wh.Register(calicoValidateWebhookPath, NewCalicoNetworkPolicyValidator(mgr.GetClient(), dec, 1001))
```
ymmt2005, line 42: ditto

admission/hooks/validate_networkpolicy_test.go
```diff
@@ -61,7 +61,7 @@ func testNewNetworkPolicy(ns, name string, order float64) runtime.Object {
 }
 
 var _ = Describe("validate networkpolicy webhook", func() {
-	It("should deny policy having order < 1000", func() {
+	It("should deny policy having order < 1001", func() {
```
ymmt2005, line 64: ditto

ymmt2005 (changes requested):

kfyharukz:
I force-pushed because commits were cluttered.

ymmt2005 (approved):
LGTM

PR #514, opened 2021-02-04
loki: run touch-protos to avoid regenerating go files for potoc

toshipp:
Signed-off-by: Toshikuni Fukaya <toshikuni-fukaya@cybozu.co.jp>


llamerada-jp (approved):
mobbed

PR #515, opened 2021-02-05
ci: omit validation with hadolint.

morimoto-cybozu:
This PR drops validation of Dockerfiles with hadolint from the CI steps.
This is because hadolint tends to report trivial or false-positive "errors."


ymmt2005 (approved):
LGTM

PR #536, opened 2021-03-04
Add scripts for not using actions-runner-controller

binoue:


tapih:
This is no longer needed due to https://github.com/cybozu-go/neco/issues/1449.

PR #529, opened 2021-02-26
Add actions-runner/actions-runner-container images

binoue:
This PR add actions-runner/actions-runner-container images.

Signed-off-by: binoue <banji-inoue@cybozu.co.jp>

PR #551, opened 2021-03-24
Purge  actions-runner/actions-runner-container images

tapih:
[This](https://github.com/cybozu/neco-containers/pull/529) is no longer needed because we develop github actions controller on our own.
Signed-off-by: Hiroshi Muraoka <h.muraoka714@gmail.com>


binoue (approved):
LGTM

PR #559, opened 2021-04-12
Update containers for Kubernetes 1.20

ysksuzuki:


argocd/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN curl -sSLf https://github.com/gobuffalo/packr/releases/download/v${PACKR_VER
 
 # Install kubectl
 # NOTE: keep the version synced with https://storage.googleapis.com/kubernetes-release/release/stable.txt
-ENV KUBECTL_VERSION=1.19.7
+ENV KUBECTL_VERSION=1.21.0
```
zoetrope, line 17: Why not 1.20?

local-pv-provisioner/Makefile
```diff
@@ -2,26 +2,44 @@
...
+test: simple-test
+ifeq (,$(wildcard $(ENVTEST_ASSETS_DIR)/setup-envtest.sh))
+	mkdir -p $(ENVTEST_ASSETS_DIR)
+	curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh $(ENVTEST_SCRIPT_URL)
+endif
+	{ \
+	source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh && \
+	fetch_envtest_tools $(ENVTEST_ASSETS_DIR) && \
+	setup_envtest_env $(PWD)/$(ENVTEST_ASSETS_DIR) && \
+	go test -v -count 1 ./... ; \
```
zoetrope, line 28: Looks like `-race` was lost.

local-pv-provisioner/controllers/persistent_volume.go
```diff
@@ -135,9 +129,7 @@ func (w *persistentVolumeWatcher) fireEvent(ctx context.Context) error {
 			continue
 		}
 		w.ch <- event.GenericEvent{
-			Meta: &metav1.ObjectMeta{
-				Name: pv.Name,
-			},
+			Object: &pv,
```
zoetrope, line 132: `pv` should not be passed as a pointer.

You can see the strange behavior in the following code.

```go
package main

import (
	"fmt"
	"time"
)

type GenericEvent struct {
	Object Object
}

type Object interface {
	GetName() string
	SetName(name string)
}

type hoge struct {
	name string
}

func(h *hoge) GetName() string{
	return h.name
}

func(h *hoge) SetName(name string) {
	h.name = name
}


func main(){
	ch := make(chan GenericEvent)

	data := []hoge{
		{
			name: "a",
		},
		{
			name: "b",
		},
		{
			name: "c",
		},
	}

	go func() {
		for _, d := range data {
			ch <- GenericEvent{
				Object:  &d,
			}
		}
	}()

	go func() {
		for {
			got := <- ch
			fmt.Println(got.Object.GetName())
		}
	}()

	time.Sleep(5*time.Second)
}
```

argocd/Dockerfile
ysksuzuki: I just followed the version [this](https://storage.googleapis.com/kubernetes-release/release/stable.txt) indicates, but 1.20 is better for now.

zoetrope (approved):
LGTM

PR #573, opened 2021-05-07
[neco-admission] mutate ephemeral-storage limit and request

zoetrope:
Replacing PSP with PSA causes `ReadOnlyRootFilesystem: false`.
As a result, there is no need to mount `/tmp` anymore.
Instead, we have to limit the size of local ephemeral storages.

### DIFF
--- a/admission/README.md
+++ b/admission/README.md
@@ -77,10 +77,11 @@ if the resource is annotated with `admission.cybozu.com/prevent: delete`.
 PodMutator
 ----------
 
-PodMutator mutates Pod manifests to mount writable emptyDir to `/tmp` for each containers.
-The purpose of this mutator is to prevent Pods from unexpected death by writing to read-only filesystem. 
+PodMutator mutates Pod manifests to specify local ephemeral storage limit to 1GiB and request to 200MiB for each container.
+The purpose of this mutator is to prevent Pods from overuse of local ephemeral storage.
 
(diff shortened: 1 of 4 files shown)


admission/README.md
```diff
@@ -77,10 +77,11 @@ if the resource is annotated with `admission.cybozu.com/prevent: delete`.
...
 ----------
 
-PodMutator mutates Pod manifests to mount writable emptyDir to `/tmp` for each containers.
-The purpose of this mutator is to prevent Pods from unexpected death by writing to read-only filesystem. 
+PodMutator mutates Pod manifests to specify local ephemeral storage limit and request for each container.
+The purpose of this mutator is to prevent Pods from overuse of local ephemeral storage.
 
-However, Pods that already have another volumes mounted under `/tmp/*` are excluded from the mutating target.
+If you want to use more ephemeral storage than the limit, you can use generic ephemeral volume instead of
+local ephemeral storage.
```
binoue, line 84: Does this need to be a new line?

Also, the request and limit of ephemeral storage for each container are fixed to a specific value.(i.e. request: 1Gi, limit: 2Gi)
It might be useful to write the information.

ymmt2005: @zoetrope @binoue 
The request value 1 Gi seems too large for me.

binoue (approved):
LGTM

PR #578, opened 2021-05-13
use latest tag for sample manifests and README.md

binoue:


ymmt2005 (changes requested):
I don't think `latest` is good enough.

Since these YAMLs are just examples for nobody (we don't need them),
why don't you simply delete YAMLs?

binoue:
`latest` tag is not recommended even in examples.
So close this PR.

PR #589, opened 2021-05-18
Update Calico to 3.19.0

morimoto-cybozu:


yokaze:
@morimoto-cybozu Is there some reason to use Go 1.15 as the builder image?

morimoto-cybozu:
@yokaze 
Yes, it's because the upstream seems to use Go 1.15.  cf. [node/go.mod](https://github.com/projectcalico/node/blob/v3.19.0/go.mod) and [typha/go.mod](https://github.com/projectcalico/typha/blob/v3.19.0/go.mod)

yokaze (approved):
LGTM

PR #601, opened 2021-05-21
Update Grafana Operator to 3.10.1

morimoto-cybozu:


grafana_plugins_init/TAG
```diff
@@ -1 +1 @@
-0.0.3.1
+0.0.3.2
```
binoue, line 1: Does this TAG change necessary?

binoue (approved):

grafana_plugins_init/TAG
morimoto-cybozu: @binoue 
Yes.
The grafana_plugins_init container image uses the Ubuntu image as its base, and we need to reflect the update of the base.
cf. https://github.com/cybozu/neco-containers/blob/update-grafana-operator-3.10.1/grafana_plugins_init/Dockerfile

PR #619, opened 2021-06-07
add s3gw

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1411


s3gw/README.md
```diff
@@ -0,0 +1,59 @@
...
+Options
+-------
+
+- `--listen`: addr:port to listen to (passed to http.Server.Addr)
+- `--use-path-style`: use path-style bucket name
+- `--hosts-allow`: subnets allowed to access to this gw, separated by comma
+- `--hosts-deny`: subnets denied to access to this gw, separated by comma
+  - note: `--hosts-allow` and `--hosts-deny` control access to `/bucket/` only. `/health` and `/metrics` accept access from everywhere.
+
+Supported API
```
ysksuzuki, line 27: Could you elaborate on the API specification such as response format, parameters if any, and support methods for `/health`, `/metrics`, and `/bucket` (Get only perhaps)? Also, It would be helpful to describe what metrics will be exported.

s3gw/allow_deny.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, fmt.Errorf("failed to parse %s CIDR: %w", name, err)
+		}
+		nets = append(nets, n)
+	}
+
+	return nets, nil
+}
+
+// ParseAllowDeny parses ACL. Note: it only accepts CIDR. (hostnames are not supported)
+func ParseAllowDeny(allowed, denied string) (*AllowDeny, error) {
```
ysksuzuki, line 34: `ParseAllowDeny` doesn't need to be exported since it is only called from the same package `main`. Or is it more appropriate to separate the package from `main`?

s3gw/handler.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		res.WriteHeader(http.StatusInternalServerError)
+		res.Write([]byte(err.Error()))
+		return
+	}
+	res.Header().Set("Content-Length", fmt.Sprint(len(marshalled)))
+	res.Write(marshalled)
+}
+
+func objectHandlerFunc(res http.ResponseWriter, req *http.Request) {
```
ysksuzuki, line 94: Let me propose to use a framework rather than implementing routing and response handling on your own. I recommend [Gin](https://github.com/gin-gonic/gin). Sorry if you have already considered this.

umezawatakeshi: It's too big for what we want to do.

s3gw/allow_deny.go
umezawatakeshi: No specific reason to capitalize.

By the way, what about the other symbols that start with a uppercase letter?

ysksuzuki (approved):

PR #623, opened 2021-06-09
Revert "[neco-admission] mutate ephemeral-storage limit and request"

zoetrope:
This reverts commit fc56ace5b815c16340b4d776d23ca7d272971531.

Unless /var/log and /var/lib/kubelet are in the same filesystem, we can't specify the limit for local ephemeral storage
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,3 @@
 .*.swp
 /.vscode
 cover.out
-vendor
-/.idea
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -4,5 +4,3 @@
 .*.swp
 /.vscode
 cover.out
-vendor
-/.idea
```
binoue, line 8: I think this part should not be reverted.

zoetrope: That's a problem with my local environment, so it's ok.

binoue (approved):

PR #633, opened 2021-06-25
rook: create a directory to make watch-active work

toshipp:
To use multiple MGR, rook attaches a sidecar container named as watch-acitive, it needs to create files under the path to work properly.
This prepares the path with sufficient owner and permission.


rook/TAG
```diff
@@ -1 +1 @@
-1.6.3.2
+1.6.3.3
```
toshipp, line 1:
```suggestion
1.6.3.4
```

llamerada-jp (approved):
LGTM

PR #638, opened 2021-07-20
Ceph: apply patch which reports error log

toshipp:
- Improve circle config by separating build shell script.
- Add a patch which reports why ListColumnFamilies is failed.


llamerada-jp (approved):
LGTM

PR #639, opened 2021-07-24
admission: allow Accurate service account to delete Namespace

ymmt2005:
Accurate is a custom Kubernetes controller to manage Namespaces.
It should be able to delete Namespaces.


zoetrope (approved):
LGTM

PR #648, opened 2021-09-02
update dctest-meows-runner

kmdkuk:
I updated the `dctest-meows-runner` image when I added it (#645), because it was missing some things to run the dctest environment on stage0.
- Change the user to execute to "actions".
  - Set `GOPATH` to `/home/actions/go`.
- ~In the image, clone neco and neco-apps (release branch). And, run `make setup` in each directory.~
- Create the `/opt/hostedtoolcache` directory for GitHubActions to use.
- update `bootstrap.sh`


dctest-meows-runner/Dockerfile
```diff
@@ -48,14 +50,22 @@ RUN apt-get update \
...
+            hub \
     && rm -rf /var/lib/apt/lists/* \
     && curl -sSLf https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /usr/local -xzf - \
     && curl -sfL https://github.com/cybozu-go/placemat/releases/download/v${PLACEMAT_VERSION}/placemat2_${PLACEMAT_VERSION}_amd64.deb -o placemat2_${PLACEMAT_VERSION}_amd64.deb \
     && dpkg -i ./placemat2_${PLACEMAT_VERSION}_amd64.deb \
     && rm ./placemat2_${PLACEMAT_VERSION}_amd64.deb \
+    && git clone -b ${NECO_BRANCH:-release} https://github.com/cybozu-go/neco.git ${NECO_DIR} \
+    && make -C ${NECO_DIR}/dctest setup \
+    && git clone -b ${NECO_APPS_BRANCH:-release} https://github.com/cybozu-go/neco-apps.git ${NECO_APPS_DIR} \
+    && make -C ${NECO_APPS_DIR}/test setup \
```
masa213f, line 62: Why do you clone here?

`neco` and `neco-apps` will be updated frequency.
Also, in the near future, they will be moved to the `cybozu-private` org and be integrated with `neco-apps-secret`.
This cloning makes risk to upload secrets to quay.io.

So I think they should be cloned at runtime.

masa213f (approved):

dctest-meows-runner/Dockerfile
masa213f, line 62: Why do you clone here?

`neco` and `neco-apps` will be updated frequency.
Also, in the near future, they will be moved to the `cybozu-private` org and be integrated with `neco-apps-secret`.
This cloning makes risk to upload secrets to quay.io.

So I think they should be cloned at runtime.

masa213f (changes requested):

dctest-meows-runner/Dockerfile
kmdkuk: It wanted to include things in the image that would be installed by `make setup`.
I think that risk certainly exists, so I will revert `clone` and `setup` to `bootstrap.sh`.

kmdkuk: fix it.

dctest-meows-runner/Dockerfile
```diff
@@ -48,14 +50,18 @@ RUN apt-get update \
...
     && rm ./placemat2_${PLACEMAT_VERSION}_amd64.deb \
     && echo "actions ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers \
-    && adduser --disabled-password --gecos "" --uid 10000 actions 
+    && adduser --disabled-password --gecos "" --uid 10000 actions \
+    && mkdir -p /opt/hostedtoolcache \
+    && chown -R actions:actions /opt/hostedtoolcache \
+    && chown -R actions:actions ${HOME}
 
-COPY bootstrap.sh /root/
+COPY bootstrap.sh ${HOME}
```
morimoto-cybozu, line 65: This COPY creates a file owned by root.  The mode of the file seems to be copied from the local environment.
If `docker build` is executed in the local copy created under `umask 007`, this COPY creates a file with mode `-rw-rw----` owned by root.  It is unreadable for actions.

```suggestion
COPY --chown=actions:actions bootstrap.sh ${HOME}
```

morimoto-cybozu (changes requested):

dctest-meows-runner/Dockerfile
kmdkuk: thank you!
I tried it in my hand, and it seems to be readable by actions without the chown option, as follows.

````
actions@a741544b98df:~$ ls -al bootstrap.sh 
-rw-rw-r-- 1 root root 815 Sep 7 02:11 bootstrap.sh
actions@a741544b98df:~$ cat bootstrap.sh 
#! /bin/sh -xe

echo "setup neco and neco-apps"
~~~~
```

However, the owner of `bootstrap.sh` is unintended, so i fixed it.

morimoto-cybozu (approved):
LGTM

masa213f (approved):

PR #649, opened 2021-09-09
upgrade golang to 1.16.7

umezawatakeshi:
Signed-off-by: UMEZAWA Takeshi <takeshi-umezawa@cybozu.co.jp>

PR #650, opened 2021-09-10
upgrade golang to 1.17.1

umezawatakeshi:


yokaze (changes requested):
I think these two files should be updated.
- `golang/analyzer/go.mod`
- `golang-bionic/analyzer/go.mod`

umezawatakeshi:
@yokaze 
It is already done in https://github.com/cybozu/neco-containers/pull/649

yokaze:
@umezawatakeshi Go version in the `go.mod` is still go 1.16. Is it OK?

umezawatakeshi:
@yokaze oops

umezawatakeshi:
I wonder that `go mod tidy` did not do so.

morimoto-cybozu:
@yokaze @umezawatakeshi 
FYI

https://golang.org/doc/go1.17#graph-pruning
> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 `go.mod` file, the newly-added requirements on _indirect_ dependencies in a `go 1.17` module are maintained in a separate `require` block from the block containing direct dependencies.

https://golang.org/issue/45965

PR #652, opened 2021-09-13
add golang-1.16

umezawatakeshi:
1.16 (1.16.7 or later) is still required by vault.


morimoto-cybozu (approved):
LGTM

PR #657, opened 2021-09-22
update admission logic for HTTPProxy in contour 1.18

umezawatakeshi:
- validation: check `.spec.ingressClassName` field
- mutation: set default ingress class to `.spec.ingressClassName` field rather than to `kubernetes.io/ingress.class` annotation


admission/hooks/mutate_httpproxy.go
```diff
@@ -43,12 +52,11 @@ func (m *contourHTTPProxyMutator) Handle(ctx context.Context, req admission.Requ
...
-
-	if ann == nil {
-		ann = make(map[string]string)
+	if _, ok, _ := getHTTPProxyIngressClassNameField(hp); ok {
+		return admission.Allowed("ok")
 	}
-	ann[annotationKubernetesIngressClass] = m.defaultClass
-	hp.SetAnnotations(ann)
+
+	setHTTPProxyIngressClassNameField(hp, m.defaultClass)
```
zoetrope, line 59: Don't we need to check for errors?

zoetrope (commented):
reviewed

admission/hooks/mutate_httpproxy.go
```diff
@@ -43,12 +52,11 @@ func (m *contourHTTPProxyMutator) Handle(ctx context.Context, req admission.Requ
 	if _, ok := ann[annotationContourIngressClass]; ok {
 		return admission.Allowed("ok")
 	}
-
-	if ann == nil {
-		ann = make(map[string]string)
+	if _, ok, _ := getHTTPProxyIngressClassNameField(hp); ok {
```
zoetrope, line 55: memo: 

> If both the annotation and the spec field are specified on an object, the annotation takes preference for backwards compatibility.
https://projectcontour.io/docs/v1.18.1/config/annotations/

morimoto-cybozu:
@umezawatakeshi 
The updated webhook does not work with Contour 1.17.  This will cause a problem because neco-admission is versioned with only one version number while it is a collection of various webhooks.
For example, if we add a new feature for PodMutator in the next version of neco-admission, and if some user updates neco-admission to use the new PodMutator while using Contour 1.17, the user will see a wrongly-defaulted HTTPProxy object.  Though the object will have a defaulted `.spec.ingressClassName` field, the field is not interpreted by Contour 1.17.

If we define neco-admission as a totally internal software, this may be acceptable.

umezawatakeshi:
So, let's define neco-admission for neco-only.

admission/Makefile
```diff
@@ -42,6 +42,12 @@ endif
 	setup_envtest_env $(PWD)/$(ENVTEST_ASSETS_DIR) && \
 	TEST_PERMISSIVE=true go test -v -count 1 ./... ; \
 	}
+	{ \
+	source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh && \
+	fetch_envtest_tools $(ENVTEST_ASSETS_DIR) && \
+	setup_envtest_env $(PWD)/$(ENVTEST_ASSETS_DIR) && \
+	NO_HTTPPROXY_MUTATION=true go test -v -count 1 ./... ; \
```
morimoto-cybozu, line 49: `NO_HTTPPROXY_MUTATION=true` can be tested at the same time with `TEST_PERMISSIVE=true`.

admission/hooks/mutate_httpproxy.go
```diff
@@ -43,12 +52,14 @@ func (m *contourHTTPProxyMutator) Handle(ctx context.Context, req admission.Requ
 	if _, ok := ann[annotationContourIngressClass]; ok {
 		return admission.Allowed("ok")
 	}
+	if _, ok, _ := getHTTPProxyIngressClassNameField(hp); ok {
```
morimoto-cybozu, line 55: Check err.

admission/hooks/mutate_httpproxy_test.go
```diff
@@ -32,25 +36,49 @@ func testHTTPProxy(name string, annotations map[string]string) map[string]string
...
-		Expect(ann).ToNot(HaveKey(annotationContourIngressClass))
+	if isHTTPProxyMutationDisabled() {
+		return
+	}
+
+	It("should have default ingress class name", func() {
+		hp := testHTTPProxy("mhp1", map[string]string{}, nil)
+		name, ok, err := unstructured.NestedString(hp.UnstructuredContent(), "spec", fieldIngressClassName)
+		Expect(ok).To(Equal(true))
+		Expect(err).ShouldNot(HaveOccurred())
```
morimoto-cybozu, line 51: Check err first.

admission/hooks/mutate_httpproxy_test.go
```diff
@@ -32,25 +36,49 @@ func testHTTPProxy(name string, annotations map[string]string) map[string]string
...
+		Expect(name).To(Equal("secured"))
 	})
 
 	It("should not mutate annotations", func() {
-		ann := testHTTPProxy("mhp2", map[string]string{annotationKubernetesIngressClass: "global"})
+		hp := testHTTPProxy("mhp2", map[string]string{annotationKubernetesIngressClass: "global"}, nil)
+		_, ok, _ := unstructured.NestedString(hp.UnstructuredContent(), "spec", fieldIngressClassName)
```
morimoto-cybozu, line 57: Check err.

admission/hooks/mutate_httpproxy_test.go
```diff
@@ -32,25 +36,49 @@ func testHTTPProxy(name string, annotations map[string]string) map[string]string
...
+		ann := hp.GetAnnotations()
 		Expect(ann).To(HaveKeyWithValue(annotationKubernetesIngressClass, "global"))
 		Expect(ann).ToNot(HaveKey(annotationContourIngressClass))
+		Expect(ok).To(BeFalse())
 	})
 
 	It("should not mutate annotations with projectcontour.io/ingress.class", func() {
-		ann := testHTTPProxy("mhp3", map[string]string{annotationContourIngressClass: "global"})
+		hp := testHTTPProxy("mhp3", map[string]string{annotationContourIngressClass: "global"}, nil)
+		_, ok, _ := unstructured.NestedString(hp.UnstructuredContent(), "spec", fieldIngressClassName)
```
morimoto-cybozu, line 66: Check err.

admission/hooks/mutate_httpproxy_test.go
```diff
@@ -32,25 +36,49 @@ func testHTTPProxy(name string, annotations map[string]string) map[string]string
...
 		Expect(ann).ToNot(HaveKey(annotationKubernetesIngressClass))
+		Expect(ok).To(BeFalse())
+	})
+
+	It("should not mutate .spec.ingressClassName field", func() {
+		ingressClassName := "global"
+		hp := testHTTPProxy("mhp4", map[string]string{}, &ingressClassName)
+		name, ok, err := unstructured.NestedString(hp.UnstructuredContent(), "spec", fieldIngressClassName)
+		Expect(ok).To(Equal(true))
+		Expect(err).ShouldNot(HaveOccurred())
```
morimoto-cybozu, line 78: Check err first.

admission/hooks/validate_httpproxy.go
```diff
@@ -30,11 +30,12 @@ func (v *contourHTTPProxyValidator) Handle(ctx context.Context, req admission.Re
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 	newAnn := hp.GetAnnotations()
+	newIngressClassNameField, _, _ := getHTTPProxyIngressClassNameField(hp)
```
morimoto-cybozu, line 33: Check err and ok.

admission/hooks/validate_httpproxy_test.go
```diff
@@ -66,8 +79,20 @@ var _ = Describe("validate HTTPProxy webhook with ", func() {
...
 	})
 
+	It("should deny httpproxy to update .spec.ingressClassName value", func() {
+		ingressClassName := "global"
+		hp := fillHTTPProxy("vhp7", map[string]string{}, &ingressClassName)
+		err := k8sClient.Create(testCtx, hp)
+		Expect(err).NotTo(HaveOccurred())
+
+		spec := hp.UnstructuredContent()["spec"].(map[string]interface{})
+		spec[fieldIngressClassName] = "forest"
```
morimoto-cybozu, line 89: `unstructured.SetNestedField()` would be usable.

morimoto-cybozu (changes requested):
Update README.

morimoto-cybozu:
Update TAG.

zoetrope (approved):
LGTM

admission/hooks/validate_httpproxy.go
umezawatakeshi: In case of annotations, nonexistent and empty value are not discriminated (= `ok` is not used). I will do so in case of `ingressClassName` field too.

admission/hooks/validate_httpproxy.go
```diff
@@ -43,12 +47,19 @@ func (v *contourHTTPProxyValidator) Handle(ctx context.Context, req admission.Re
 			return admission.Errored(http.StatusBadRequest, err)
 		}
 		oldAnn := old.GetAnnotations()
+		oldIngressClassNameField, _, err := getHTTPProxyIngressClassNameField(old)
```
morimoto-cybozu, line 50: Check ok.

morimoto-cybozu (changes requested):

admission/hooks/validate_httpproxy.go
morimoto-cybozu: It's OK to equate `newIngressClassNameField == ""` to `!ok`.
What I mean is that there is no guarantee `newIngressClassNameField` is equal to `""` when `ok` is false.

In case of maps, the Go language guarantees that.
https://golang.org/doc/effective_go#maps
> An attempt to fetch a map value with a key that is not present in the map will return the zero value for the type of the entries in the map.

morimoto-cybozu:
As commented above, it's OK to equate `newIngressClassNameField == ""` to `!ok`.
There is no guarantee `newIngressClassNameField` is equal to `""` when `ok` is false.

morimoto-cybozu:
> As commented above, it's OK to equate `newIngressClassNameField == ""` to `!ok`.

This is misleading...
What I mean is that it's ok to assign `newIngressClassNameField = ""` and forget about `ok` after confirming `!ok`.

morimoto-cybozu (approved):
LGTM

PR #672, opened 2021-10-12
add vmbackup/vmrestore/vmctl container images

umezawatakeshi:


morimoto-cybozu:
@umezawatakeshi 
If vmbackup/vmrestore/vmctl are executed not as long-term daemons but as short-term tools, we could provide them in a single container image.
I assume that dividing VictoriaMetrics binaries into several container images is to minimize long-term Pods' footprints.  Is my understanding correct at all?

umezawatakeshi:
I just followed the official images construction. There are individual images for each program. (e.g. https://hub.docker.com/r/victoriametrics/vmbackup ).

> I assume that dividing VictoriaMetrics binaries into several container images is to minimize long-term Pods' footprints. Is my understanding correct at all?

I think you are right.

PR #678, opened 2021-10-19
[neco-admission] Resources in dev-namespaces ignore DeleteValidator

zoetrope:


yokaze:
@zoetrope 
I could not figure out the reason to prevent `ResourceQuota` from deletion.
I suspect it is added to test a namespaced resource, is it true?

PR #697, opened 2021-11-19
Add cAdvisor

morimoto-cybozu:


maintenance.md
```diff
@@ -199,8 +212,7 @@ In Regular update, do the following as part of the update of each CRD-providing
 2. Check the upstream Dockerfile. If there are any updates, update our `Dockerfile`.
    - https://github.com/jimmidyson/configmap-reload/blob/vX.Y.Z/Dockerfile
 3. Update `CONFIGMAP_RELOAD_VERSION` in `Dockerfile`
-4. Update `src/CHANGELOG.md`.
-5. Update `BRANCH` and `TAG` files.
+4. Update `BRANCH` and `TAG` files.
```
ymmt2005, line 215: Why don't we maintain only `TAG` file?

morimoto-cybozu: It would be just a convention.
I've never seen the use of branches for container image tags, especially as to `0` branch.

So I'll remove `BRANCH` for cAdvisor.
For other containers, the regular update would be the right time.

ymmt2005 (approved):
LGTM

PR #709, opened 2021-12-14
Add iproute2 to dctest-meows-runner

kmdkuk:


ysksuzuki (approved):

dctest-meows-runner/Dockerfile
```diff
@@ -47,6 +47,7 @@ RUN apt-get update \
             time \
             kmod \
             iptables \
+            iproute2 \
             openssh-client \
             sudo \
             hub \
```
ymmt2005, line 53: unrelated, but `hub` will not be used.

kmdkuk: Thank you.
I also tried to run the job of creating PR with neco-tenant-apps on this image and forgot to delete it.
Indeed, this is not used now, so I will delete it too.

ymmt2005: If we really want this, I'd suggest `gh` instead of `hub`.

ymmt2005 (approved):
LGTM

PR #715, opened 2021-12-15
machines-endpoints: take care of the case when Endpoints/EndpointSlice do not exist

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1858


masa213f (approved):
Endpoint resource can be created by a PUT. So the creation of endpoint is not necessary.
https://github.com/kubernetes/kubernetes/blob/v1.22.4/pkg/registry/core/endpoint/strategy.go#L67-L70

However, it's easier to understand now.

PR #729, opened 2021-12-28
Run hubble-relay as root

ysksuzuki:
hubble-relay needs to run as root since it communicates with cilium-agent using a unix domain socket `/var/run/cilium/hubble.sock` on host. This requires root privilege.


morimoto-cybozu (approved):
LGTM

PR #759, opened 2022-03-03
add argocd-notifications v1.2.1

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/778


argocd-notifications/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+FROM quay.io/cybozu/golang:1.17-focal as builder
+
+ENV REPO=argoproj-labs/argocd-notifications
+ENV VERSION=1.2.1
+
+RUN apt-get update && apt-get install ca-certificates
+
+WORKDIR /work
+RUN curl -sSLf "https://github.com/${REPO}/archive/v${VERSION}.tar.gz" | \
```
morimoto-cybozu, line 11: It's better to place `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` when using a pipeline to catch an error at the left-side of the pipe.

Even without it, the right-side "tar" would notice an corrupted input and fail, though.

argocd-notifications/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    tar --strip-components=1 -zxf -
+RUN go mod download
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /app/argocd-notifications ./cmd
+RUN ln -s /app/argocd-notifications /app/argocd-notifications-backend
```
morimoto-cybozu, line 15: This creates a symbolic link in the builder container, but the COPY below creates a regular file.
This would not be what you want to do.

Executing `RUN ln -s` in the target container should work well.

morimoto-cybozu (changes requested):
Though I know that the argocd-notifications container is transient, I'd like to point out the common pitfalls in Dockerfile.

morimoto-cybozu:
@umezawatakeshi 
Oops, we cannot execute `RUN` on a scratch container because it does not have a shell binary...

PR #769, opened 2022-03-18
 add cephcsi container.

peng225:


.circleci/config.yml
```diff
@@ -150,8 +150,9 @@ jobs:
       - run:
           name: Check TAG file
           command: |
-            c="$(./tag_exists ceph)"
-            if [ "$c" = ng ]; then
+            ceph_tag_exists="$(./tag_exists ceph)"
+            cephcsi_tag_exists="$(./tag_exists cephcsi)"
+            if [ "$ceph_tag_exists" = ng -o "$cephcsi_tag_exists" = ng ]; then
```
toshipp, line 155: I prefer using `||` operator instead of `-o`.
See manpage.
>NOTE: Binary -a and -o are inherently ambiguous.  Use 'test EXPR1 && test EXPR2' or  'test
>      EXPR1 || test EXPR2' instead.

https://manpages.ubuntu.com/manpages/hirsute/en/man1/test.1.html

toshipp: BTW, why ceph build is needed when cephcsi image is not found?

cephcsi/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+ARG GO_ARCH
+ARG SRC_DIR
+ARG GIT_COMMIT
+ARG GOROOT=/usr/local/go
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+RUN apt-get update && apt-get -y install \
+    git \
+    && true
```
toshipp, line 21: Why it is needed? Did you copy this from upstream?
https://github.com/ceph/ceph-csi/blob/4652b8facfa56f6619967f1466b548bd6a924f19/deploy/cephcsi/image/Dockerfile#L38

toshipp (changes requested):

cephcsi/Dockerfile
peng225: >Did you copy this from upstream?

Exactly. But, honestly, I was not sure understanding why it is needed.
According to the following document, it is to prevent editing errors.
https://github.com/projectatomic/container-best-practices/blob/master/creating/creating_index.adoc
>Editing a line with a command that that you don’t want to change presents a risk of introducing a bug, and you also obscure the line’s history. This can be mitigated by having both the first and last commands true - they don’t do anything.

In our case, only git should be installed here for `git clone`, and it is unlikely that other programs are added (other programs may be added to the parameter list of another `apt-get install` command appearing later). So I removed it.

.circleci/config.yml
peng225: >I prefer using || operator instead of -o.

I replaced `-o` with `||`.

peng225: >BTW, why ceph build is needed when cephcsi image is not found?

That is the most complicated point.
Shortly speaking, Ceph build is required to resolve the dependencies between ceph-related packages.

First of all, the upstream cephcsi container is built from the ceph official image.
To completely stop depending on the container registries hosted by other organizations than Cybozu, our cephcsi container should be created from our self-built ceph image.
However, because our ceph image contains a relatively new `librados2` package and installation of the `librados-dev` package via `apt-get` requires an older version of `librados2`, the installation of `librados-dev` fails. To resolve this problem, I came up with installing librados-dev from deb package which is generated by ceph build.

cephcsi/Dockerfile
toshipp: Got it

.circleci/config.yml
toshipp: Understood.

toshipp (approved):
LGTM

cupnes (approved):

PR #787, opened 2022-04-26
ceph: set bottom most level compaction to kForce.

peng225:


peng225:
This PR was created to resolve the performance issue of RGW, but it was resolved by upgrading Ceph version. So I close this PR.

PR #797, opened 2022-05-13
update unbound_exporter to support SO_REUSEPORT

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1975

patch is https://github.com/cybozu-go/unbound_exporter/commit/01b7d834bd32fe6f3806b06b368d7c0a90b5cb3b


masa213f:
Comment to https://github.com/cybozu-go/unbound_exporter/commit/01b7d834bd32fe6f3806b06b368d7c0a90b5cb3b

The `http.ListenAndServe` uses a blank `ListenConfig`.

https://github.com/golang/go/blob/go1.17.10/src/net/http/server.go#L3183-L3186
> ```go
> func ListenAndServe(addr string, handler Handler) error {
> 	server := &Server{Addr: addr, Handler: handler}
> 	return server.ListenAndServe()
> }
> ```
https://github.com/golang/go/blob/go1.17.10/src/net/http/server.go#L2919-L2932
> ```go
> func (srv *Server) ListenAndServe() error {
> 	if srv.shuttingDown() {
> 		return ErrServerClosed
> 	}
> 	addr := srv.Addr
> 	if addr == "" {
> 		addr = ":http"
> 	}
> 	ln, err := net.Listen("tcp", addr)[Brad Fitzpatrick, 8 years ago: • net/http: use TCP keep-alives for ListenAndSe…](https://sourcegraph.com/github.com/golang/go/-/commit/d6bce32a3607222075734bf4363ca3fea02ea1e5)
> 	if err != nil {
> 		return err
> 	}
> 	return srv.Serve(ln)
> }
> ```
https://github.com/golang/go/blob/go1.17.10/src/net/dial.go#L710-L713
> ```go
> func Listen(network, address string) (Listener, error) {
> 	var lc ListenConfig
> 	return lc.Listen(context.Background(), network, address)
> }
> ```

So I think it is better to set the `ListenConfig.Control` only when the `web.reuse-port` is set.

```go
	_ = level.Info(log).Log("Listening on address:port => ", *listenAddress)
	lc := net.ListenConfig{}
	if *reusePort {
		lc.Control = func(network, address string, c syscall.RawConn) error {
			var operr error
			fn := func(s uintptr) {
				operr = unix.SetsockoptInt(int(s), unix.SOL_SOCKET, unix.SO_REUSEPORT, 1)
			}
			err := c.Control(fn)
			if err != nil {
				return err
			}
			return operr
		}
	}
	l, err := lc.Listen(context.Background(), "tcp", *listenAddress)
	_ = level.Error(log).Log(err)
	if err != nil {
		os.Exit(1)
	}
	_ = level.Error(log).Log(http.Serve(l, nil))
	os.Exit(1)
```

masa213f (approved):

PR #821, opened 2022-07-07
ceph: build ceph with dev libraries

toshipp:
cephcsi requires dev ceph libraries, but its base image does not
contain them. When building cephcsi, it reuses artifacts build by ceph
job. If ceph is updated but cephcsi uses older base image, version
mismatch will happen.
To avoid such problem, we now make ceph image include dev libraries.
Also, add a job to build old ceph image to build cephcsi temporarily.

issue: https://github.com/cybozu/csa/issues/96


toshipp:
we should build new ceph image that contains dev libraries.

ceph/build.sh
```diff
@@ -53,7 +53,6 @@ rm debian/libcephfs-java.jlibs debian/libcephfs-jni.install debian/ceph-mgr-dash
 # To avoid OOM killer, use 10 parallelism instead of 20 (max vCPU).
 dpkg-buildpackage --build=binary -uc -us -j10
 rm ../*-dbg_*.deb
-mv ../*-dev_*.deb ../workspace/dev/
```
peng225, line 56: Now you do not need to exec `mkdir -p src/workspace/dev/` above.

.circleci/config.yml
```diff
@@ -170,6 +170,38 @@ jobs:
...
+    docker:
+      - image: quay.io/cybozu/ubuntu-dev:20.04
+    resource_class: 2xlarge+
+    steps:
+      - checkout
+      - run:
+          name: Check TAG file
+          command: |
+            ceph_tag_exists="$(./tag_exists ceph ceph-16 -dev)"
+            cephcsi_tag_exists="$(./tag_exists cephcsi)"
```
toshipp, line 186: we have no need to build ceph if it is necessary to build cephcsi, can delete this.

.circleci/config.yml
```diff
@@ -660,7 +702,7 @@ workflows:
           attach: true
           dir: cephcsi
           requires:
-            - build-ceph
+            - build-ceph-16
```
toshipp, line 705: Note: In this PR, we build ceph before cephcsi, the requires is needed.
It can be deleted after this is merged.

peng225 (approved):

PR #825, opened 2022-07-11
Add patch 19451 and 19800 for Cilium

ysksuzuki:
This commit adds patches 19451 and 19800 for cilium-agent container. Those patches partially resolve the issue that cilium sends packets to inactive backends if clients hit old connections that have been already closed. Cilium selects new backends for syn packets if they hit the same old connections that have been closed for more than 30s.

See
https://github.com/cilium/cilium/pull/19451
https://github.com/cilium/cilium/pull/19800

Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>


Hsn723 (approved):

zoetrope (approved):
LGTM

PR #834, opened 2022-07-29
update images

masa213f:


morimoto-cybozu (changes requested):

pause/Dockerfile
```diff
@@ -3,8 +3,8 @@
 # Stage1: build from source
 FROM quay.io/cybozu/ubuntu-dev:20.04 AS build
 
-ARG K8S_VERSION=1.23.0
-ARG PAUSE_VERSION=3.6
+ARG K8S_VERSION=1.24.3
+ARG PAUSE_VERSION=3.7
```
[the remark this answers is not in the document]
masa213f: I see. I will update `K8S_VERSION` only.

The `pause` is scratch-based image and the source code has not been changed.
So I think we need not rebuild it. What do you think about it?

morimoto-cybozu (approved):
LGTM

etcd/TAG
```diff
@@ -1 +1 @@
-3.5.4.1
```
kmdkuk, line 1: I didn't think it was necessary to update the tag if it was just a README modification.

masa213f: The etcd image is a ubuntu-based image. So I intent to update the base image.

kmdkuk (approved):
LGTM

PR #837, opened 2022-08-08
Install openssh-client on Argo CD image

yokaze:


argocd/Dockerfile
```diff
@@ -55,7 +55,7 @@ RUN groupadd -g 10000 argocd && \
     mkdir -p /home/argocd && \
     chown argocd:argocd /home/argocd && \
     apt-get update && \
-    apt-get install -y --no-install-recommends git gpg gpg-agent && \
```
zoetrope, line 58: ArgoCD's Dockerfile does not have openssh-client. (Probably openssh-server contains openssh-client)
https://github.com/argoproj/argo-cd/blob/master/Dockerfile

Do you make sure this works?

yokaze: @zoetrope I thought it would work, but I didn't check it. I'll briefly check it on a kind cluster. Thank you!

yokaze: This image can load manifests from a private repository with a deploy key.

zoetrope (approved):
LGTM

PR #845, opened 2022-08-25
admission: add DeploymentReplicaCountValidator

daichimukai:
This validator enforces that the number of replicas for a Deployment is
0 if the Deployment has a specific annotation.  This may be useful to
prevent the number of replicas from being rewritten by ArgoCD or
operators when the number of replicas is intentionally set to 0.


peng225 (approved):

daichimukai:
need to bump version

peng225 (approved):

admission/hooks/validate_deployment_replica_count_test.go
```diff
@@ -0,0 +1,146 @@
...
+kind: Deployment
+metadata:
+  name: ubuntu-deployment-4
+  namespace: default
+  labels:
+    app: ubuntu
+  annotations:
+    admission.cybozu.com/force-replica-count: "2"
+spec:
+  replicas: 1
```
morimoto-cybozu, line 124: It would be good to clarify that this Deployment is erroneous even though force-replica-count == .spec.replicas.

```suggestion
    admission.cybozu.com/force-replica-count: "1"
spec:
  replicas: 1
```

morimoto-cybozu (changes requested):

admission/hooks/validate_deployment_replica_count_test.go
daichimukai: Changed to test both cases of `force-replica-count == .spec.replicas.` and `force-replica-count != .spec.replicas.`.

daichimukai:
We've rewritten the test with the table test.

admission/hooks/validate_deployment_replica_count_test.go
```diff
@@ -0,0 +1,127 @@
...
+			hasAnnotation:     true,
+			forceReplicaCount: 1,
+			expectAccepted:    true,
+		},
+	}
+
+	for i, tt := range testcases {
+		manifestName := fmt.Sprintf("ubuntu-deployment-%d", i)
+
+		It(tt.scenario, func() {
```
daichimukai, line 106: We may need `tt := tt` before here, otherwise last `tt` is always used.

morimoto-cybozu (approved):
LGTM

peng225 (approved):

PR #850, opened 2022-08-30
Install jq on Registry Container Image

zeroalphat:
When doing GC(Garbage Collection), it is necessary to access the API to check if the container image exists.
Since GC are contained in the registry image, we need to add `jq` to the registry image to parse the API results.

Signed-off-by: Taichi Takemura [taichi-takemura@cybozu.co.jp](mailto:taichi-takemura@cybozu.co.jp)


registry/Dockerfile
```diff
@@ -18,6 +18,9 @@ RUN GO111MODULE=off CGO_ENABLED=0 make clean binaries VERSION=${REGISTRY_VERSION
 # Stage2: setup runtime container
 FROM quay.io/cybozu/ubuntu:20.04
 
+RUN apt-get update && apt-get install -y \
+    jq
```
morimoto-cybozu, line 22: You can shrink the container image by removing package list files created by `apt-get update`.
In my own test, the image was downsized from 172 MB to 135 MB.
cf. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get
cf. https://github.com/cybozu/neco-containers/blob/main/bird/Dockerfile and other Dockerfiles

Note:
* `apt-get clean` is automatically executed
* `--no-install-recommends` is also useful in general; it makes no difference for `jq`, though
* The cleanup is meaningful only for the final stage; it does not have meaning for intermediate containers

```suggestion
RUN apt-get update && apt-get install -y --no-install-recommends \
    jq && \
    rm -rf /var/lib/apt/lists/*
```

morimoto-cybozu (changes requested):

registry/Dockerfile
zeroalphat: Thank you for the review.
Reduce container image at 68309c324b29fa94532b777c6b9d33de785f9e06.

morimoto-cybozu (approved):
LGTM

morimoto-cybozu (approved):

PR #865, opened 2022-09-22
Use (*http.Transport).Clone() instead of copying http.DefaultTransport manually

arosh:
Copying http.DefaultTransport manually is troublesome to follow if properties of http.Transport vary with upgrading of Go.


yokaze (changes requested):
Sorry for the super-late review.
I confirmed the changes are valid, could you please rebase it and reassign new version numbers?

yokaze (approved):
LGTM

PR #868, opened 2022-09-26
Validate that an ephemeral container has a valid image

zoetrope:
Current neco-admission does not validate EphemeralContainers.
This PR supports EphemeralContainers as the webhook targets.


admission/hooks/validate_pod_test.go
```diff
@@ -90,4 +92,40 @@ spec:
...
+				Name:  "debug",
+				Image: "busybox",
+			},
+		}
+		po.Spec.EphemeralContainers = append(po.Spec.EphemeralContainers, ephemeralContainer)
+
+		k8s, err := kubernetes.NewForConfig(k8sConfig)
+		Expect(err).NotTo(HaveOccurred())
+
+		po, err = k8s.CoreV1().Pods(po.Namespace).UpdateEphemeralContainers(testCtx, po.Name, po, metav1.UpdateOptions{})
```
morimoto-cybozu, line 127: `statikcheck` complains.

```
hooks/validate_pod_test.go:127:3: this value of po is never used (SA4006)
```

morimoto-cybozu (changes requested):

morimoto-cybozu (changes requested):
Please update `admission/cmd/run.go` too.
cf. https://github.com/cybozu/neco-containers/pull/853

This is a common pitfall.
I also have overlooked this type of error.
It would be nice to document which files we should edit.

admission/hooks/validate_pod_test.go
```diff
@@ -90,4 +92,40 @@ spec:
...
+			},
+		}
+		po.Spec.EphemeralContainers = append(po.Spec.EphemeralContainers, ephemeralContainer)
+
+		k8s, err := kubernetes.NewForConfig(k8sConfig)
+		Expect(err).NotTo(HaveOccurred())
+
+		_, err = k8s.CoreV1().Pods(po.Namespace).UpdateEphemeralContainers(testCtx, po.Name, po, metav1.UpdateOptions{})
+		permissive := os.Getenv("TEST_PERMISSIVE") == "true"
+		Expect(err == nil).To(Equal(permissive))
```
zeroalphat, line 129: Is it correct that the meaning of the code here is that it confirms that err is nil?
I ask because I thought that if adding an ephemeral container fails due to admission webhook deny, err would not be nil.

zoetrope: When PERMISSIVE mode is enabled, err will not be nil.
However, this test is difficult to understand. I will rewrite it.

zoetrope: https://github.com/cybozu/neco-containers/pull/868/commits/0670a214928f09be5adb97d59518c89324763744

morimoto-cybozu:
@zoetrope 
Oh, I see.
Well, do we need `ephemeralContainerValidatingWebhookPath` in `admission/hooks/suite_test.go`?

zoetrope:
@morimoto-cybozu 
Thanks. Removed it.

morimoto-cybozu (approved):
LGTM

admission/hooks/validate_pod_test.go
zeroalphat: Thank you for rewriting the code.
It helped me to understand it.

zeroalphat (approved):
LGTM

PR #879, opened 2022-10-05
Install CA certificates to blackbox-exporter

takahisa:
In this PR, I copied the certificates included in the `ca-certificates` package installed at the build stage to the appropriate directory in the scratch image. 

Since the `quay.io/repository/cybozu/blackbox_exporter` image uses scratch as the base image, the CA certificate was not installed. Therefore, monitoring endpoints over HTTPS cannot be performed from blackbox-exporter. This PR resolves this monitoring issue.


blackbox_exporter/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN curl -fsSL https://github.com/prometheus/blackbox_exporter/archive/v${BLACKB
 RUN make build
 
 # Stage2: setup runtime container
-FROM scratch
+FROM quay.io/cybozu/ubuntu:20.04
```
Hsn723, line 15: [IMO] This is an overkill solution to the initial problem which was the lack of installed CA certificates. Using the Ubuntu base image increases the attack surface. With Ubuntu, this container may need to be rebuilt more frequently due to vulnerabilities in unrelated Ubuntu packages, whereas a simple scratch + ca-certificates would not require such attention.

takahisa: @Hsn723 
Please see NECOREQ-205 on Kintone.
the idea of copying the CA certificate put in the build stage into the scratch image has been canceled after discussion.

masa213f: Thank you. I prefer a simple solution.
Our ubuntu image is not complicated, so I don't think we need to worry about the secury.

masa213f (dismissed):
LGTM

PR #918, opened 2022-12-14
Match the versions in sample manifests

yokaze:
This is the minor follow-up of https://github.com/cybozu/neco-containers/pull/865.


terassyi (approved):
LGTM

PR #923, opened 2022-12-19
Update containers held by the monitoring team

kmdkuk:
- The following container wants to create an Image with only the base image updated, so only the TAG file is updated
  - alertmanager
  - heartbeat
- The following container updates the version of meows used for base
  - meows-dctest-runner
  - meows-neco-runner
 

masa213f (approved):
LGTM

PR #929, opened 2023-01-10
Use net.JoinHostPort() instead of fmt.Sprintf()

arosh:
See https://blog.lufia.org/entry/2022/12/16/205728


masa213f (approved):
LGTM. Thank you!

PR #931, opened 2023-01-12
Update container image tag stable to current-20.04

zeroalphat:


morimoto-cybozu (approved):
LGTM

PR #962, opened 2023-02-08
admission: update envtest k8s version

zeroalphat:
Update `ENVTEST_K8S_VERSION` to 1.25.0.


yokaze (changes requested):
Could you please write in `maintenance.md` that we need to update `ENVTEST_K8S_VERSION`?
https://github.com/cybozu/neco-containers/blob/main/maintenance.md#admission-neco-admission

yokaze (approved):
LGTM

PR #964, opened 2023-02-08
pause: update for k8s 1.25

zeroalphat:


yokaze (changes requested):
Could you please write in `maintenance.md` that we need to update `K8S_VERSION`?
https://github.com/cybozu/neco-containers/blob/main/maintenance.md#pause

yokaze (approved):
LGTM

PR #970, opened 2023-02-13
update teleport-node to 11.3.3

terassyi:
We need nodejs and yarn command to build teleport-node.


maintenance.md
```diff
@@ -804,8 +804,10 @@ Only the base image and module dependency should be updated.
 2. Check the upstream `Makefile` and `docker/Dockerfile`. If they have been updated significantly, update our `Dockerfile`.
    - https://github.com/gravitational/teleport/blob/vX.Y.Z/Makefile
    - https://github.com/gravitational/teleport/blob/vX.Y.Z/docker/Dockerfile
-3. Update `TELEPORT_VERSION` in `Dockerfile`.
-4. Update `BRANCH` and `TAG` files.
+3. Update `NODE_VERSION` in `Dockerfile`. Pleas reference bellow.
```
yokaze, line 807: It would be better to write the exact instruction.

```suggestion
3. Update `NODE_VERSION` in `Dockerfile`. Please use the version in the below file.
```

yokaze (changes requested):
Please remove the `hoge` file and follow the comment for `NODE_VERSION`.

maintenance.md
terassyi: Thank you for your comment.
I fixed it.

yokaze (approved):
LGTM

PR #975, opened 2023-02-15
Add MySQL 8.0.32 image

masa213f:
- Restore files for the MySQL 8.0.32 image, which deleted in https://github.com/cybozu/neco-containers/pull/974.
- Rebuild mysql 8.0.18～8.0.31 images.


mysql/8.0.32/TAG
```diff
@@ -0,0 +1 @@
+8.0.32.1
```
umezawatakeshi, line 1: should be 8.0.32.<b>2</b> in this case, isn't it?

masa213f: `mysql:8.0.32.1` is not broken and still exist on quay.io. So it need not rebuild.

umezawatakeshi (approved):

PR #1013, opened 2023-04-14
Fix squid and cilium images

ysksuzuki:
- Revert dependencies for cilium-image-tools
  - cilium-agent uses ubuntu 20.04 base image, so we need to stick to the 20.04 dependencies for image- tools
- Add `disable-arch-native` flag for squid build to fix the `Illegal instruction (core dumped)` error
  - The default is to auto-detect compiler support and use where available, but it leads to unsupported optimization in the running environment.


morimoto-cybozu (approved):
LGTM

PR #1014, opened 2023-04-17
Add dependencies for Cilium runtime image

ysksuzuki:
Ref:
https://github.com/cilium/cilium/blob/4e3c3c5bc790a410bf1f31757f30259e6dbba454/images/runtime/Dockerfile#L41
https://github.com/cilium/cilium/blob/4e3c3c5bc790a410bf1f31757f30259e6dbba454/images/runtime/install-runtime-deps.sh#L11-L22


morimoto-cybozu (approved):
LGTM

PR #1024, opened 2023-05-15
Grafana v9.5.2

masa213f:
The `quay.io/cybozu/grafana:9.4.3.1` image is wrong.

From Grafana v9.4.0, `run.sh` executes the `grafana` binary instead of the `grafana-server`.
But our grafana image does not contain the `grafana` binary.

https://github.com/grafana/grafana/blob/v9.4.0/packaging/docker/run.sh#L80

So I update the Dockerfile.
In addition, I update the Grafana to the latest version.


grafana/Dockerfile
```diff
@@ -23,10 +23,10 @@ ENV NODE_ENV production
 RUN yarn build
 
 RUN go mod verify
-RUN make build-go
+RUN make build-go # GO_BUILD_TAGS=${GO_BUILD_TAGS} WIRE_TAGS=${WIRE_TAGS}
```
yamatcha, line 26: Just Question.
What's the purpose of this comment.
I can't find description like this in upstream build procedure.

masa213f: Thank you. I forgot to check this tags which seems be added v9.5.0
https://github.com/grafana/grafana/blob/v9.5.0/Dockerfile#L67

yamatcha: Thank you for your answer. I understand.
I was looking the v9.4.0 Dockefile.

yamatcha (dismissed):

grafana/Dockerfile
masa213f: I confirmed it. We need not set these options, so I removed them.

yamatcha (approved):

PR #1034, opened 2023-05-31
Add option to specify arm64 at container build 

zeroalphat:
Add option to build container image for arm64.
Container images were built for amd64 and arm64 in the past, now by default they will only be built for amd64.


morimoto-cybozu (approved):
LGTM

PR #1050, opened 2023-06-28
Update container images for k8s 1.26

yokaze:
Update container images for k8s v1.26 except for kubernetes itself.


unbound_exporter/TAG
```diff
@@ -1 +1 @@
-0.4.1.5
+0.4.1.6
```
masa213f, line 1: unbound_exporter is a scratch-based image.
If the upstream has not been updated, so we need no rebuild the image.

masa213f (changes requested):

masa213f (approved):

PR #1065, opened 2023-07-27
Update bird, cert-manager, and chrony

yokaze:
bird:
https://github.com/cybozu/neco-containers/blob/main/maintenance.md#bird

cert-manager:
https://github.com/cybozu/neco-containers/blob/main/maintenance.md#cert-manager

chrony: update base image only


terassyi (approved):
LGTM

PR #1069, opened 2023-08-08
update registry to 2.8.2

YZ775:


registry/Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Stage1: build from source
 FROM quay.io/cybozu/golang:1.19-jammy AS build
```
zeroalphat, line 4: Please update the version of golang as well.
```suggestion
FROM quay.io/cybozu/golang:1.20-jammy AS build
```

zeroalphat (changes requested):

zeroalphat (approved):
LGTM

PR #1073, opened 2023-08-17
delete kube-summary-exporter

YZ775:
We decided not to use kube-summary-exporter.
This PR deletes kube-summary-exporter.


zeroalphat (approved):
LGTM

PR #1076, opened 2023-08-23
update argocd to 2.8.1

YZ775:
- Update ArgoCD to 2.8.1
  - adopt tini to avoid PID1 problem
- Update dex to 2.37.0
- Update haproxy to 2.6.14


zoetrope (approved):
LGTM

PR #1080, opened 2023-08-31
add option to not run radosgw-admin

TakabayaP:
added option -norgw to disable the feature collecting metrics from radosgw-admin which creates unnecessary RGW related pools for some reason


daichimukai:
I think `--norgw=false` is confusing since it has two negations in one option. Alternatively, we can create an option like `--export-rgw-metrics` which enables a function to export rgw related metrics. What do you think?

TakabayaP:
I wanted to make sure that the command will work the same way it did without options, so my idea was to set -norgw like `ceph-extra-exporter -norgw`. However, I agree that the `-norgw` option is a bit confusing since it has negations, so making an option like `--export-rgw-metrics` and setting it's default value to true seems like a better idea!

TakabayaP:
Changed from `-norgw` (`false` by default) to `--export-rgw-metrics` (`true` by default).

cupnes (commented):
Please update the version of the `ceph-extra-exporter/TAG` file. I think it is 0.1.6 by updating the patch version.

peng225:
@cupnes @TakabayaP 
> I think it is 0.1.6 by updating the patch version.

Because a command line option (which is visible to users) is added, I think we should update the minor version.
(Though there is no clear rule of the versioning policy when the major version is 0...)

peng225 (approved):

TakabayaP:
Incrementing the minor version seems more like it to me, so decided to go with v0.2.0.
Completely forgot to change the versioning, thank you very much!

ceph-extra-exporter/TAG
```diff
@@ -1 +1 @@
-0.1.5
+0.2.0
```
peng225, line 1: Please add a new line at the end of file.

peng225 (approved):

cupnes (approved):

PR #1086, opened 2023-09-14
Update cilium to v1.13.7 & update hubble to v0.11

terassyi:
Update following components...
- cilium
- cilium-certgen
- cilium-operator-generic
- hubble
- hubble-relay
- hubble-ui

Also, cilium is patched geneve DSR mode and RateLimiter Latency metrics changes.


cilium/Dockerfile
```diff
@@ -1,20 +1,20 @@
...
+ARG UBUNTU_IMAGE=quay.io/cybozu/ubuntu:22.04
 ARG DESTDIR=/tmp/install/linux/amd64
 
 # Stage1: build common
 FROM ${GOLANG_IMAGE} as build-base
 ARG DESTDIR
 ENV DESTDIR=${DESTDIR}
 COPY TAG /
-COPY 18449.patch /tmp/
+COPY cilium.patch /tmp/
```
ysksuzuki, line 10: I think we can clone our fork repository, including the patches, instead of patching the original repository
`git clone --depth 1 --branch <BRANCH> https://github.com/cybozu-go/cilium`

.circleci/config.yml
```diff
@@ -486,14 +486,14 @@ jobs:
...
             cat > /etc/apt/sources.list \<< EOF
-            deb [arch=amd64] http://archive.ubuntu.com/ubuntu focal main restricted universe multiverse
-            deb [arch=amd64] http://security.ubuntu.com/ubuntu focal-updates main restricted universe multiverse
-            deb [arch=amd64] http://security.ubuntu.com/ubuntu focal-security main restricted universe multiverse
-            deb [arch=amd64] http://archive.ubuntu.com/ubuntu focal-backports main restricted universe multiverse
-            deb [arch=arm64] http://ports.ubuntu.com/ focal main restricted universe multiverse
-            deb [arch=arm64] http://ports.ubuntu.com/ focal-updates main restricted universe multiverse
-            deb [arch=arm64] http://ports.ubuntu.com/ focal-security main restricted universe multiverse
-            deb [arch=arm64] http://ports.ubuntu.com/ focal-backports main restricted universe multiverse
+            deb [arch=amd64] http://archive.ubuntu.com/ubuntu jammy main restricted universe multiverse
```
ysksuzuki, line 489: I see some divergence in the dependencies, e.g. `gcc-9`. Should we fix those?
https://github.com/cilium/image-tools/blob/ff22ba3bff1010f4a2dd76ede789663c3beaf8d2/images/compilers/install-deps.sh

hubble-ui/BRANCH
```diff
@@ -1 +1 @@
-0.10
+0.12
```
ysksuzuki, line 1: The default hubble-ui version in the Cilium v1.13.7 is 0.11. I would suggest using hubble-ui v0.11 for now.
https://github.com/cilium/cilium/blob/1.13.7/install/kubernetes/cilium/values.yaml#L1225

cilium/TAG
```diff
@@ -1 +1 @@
-1.12.11.1
+1.13.6.1
```
ysksuzuki, line 1: The latest for v1.13 is 1.13.7. Can't we use v1.13.7?

cilium-certgen/TAG
```diff
@@ -1 +1 @@
-0.1.8.1
+0.1.9.1
```
ysksuzuki, line 1: Cilium v1.13.6 still uses certgen 0.1.8 by default. I think we should use v1.13.7 with certgen v0.1.9
https://github.com/cilium/cilium/blob/1.13.6/install/kubernetes/cilium/values.yaml#L843

ysksuzuki (changes requested):

hubble-ui/BRANCH
terassyi: Thank you for your review.

In this document, hubble v0.11 is not maintained now.
And hubble v0.12 supports cilium v0.14 and older version.
So I think we should use hubble v0.12 and the same version of hubble-ui.

https://github.com/cilium/hubble#releases

ysksuzuki: This document is about hubble cli, not hubble-ui, right? As far as I know, the versions of hubbl cli and hubble-ui do not have to the same.

The reason I suggest it is that I used the latest hubble-ui that Cilium hasn't supported and that broke it due to the Nginx config change.

ysksuzuki: FYI, nginx.conf for hubble-ui is here
https://github.com/cilium/cilium/tree/main/install/kubernetes/cilium/templates/hubble-ui

ysksuzuki (approved):

PR #1091, opened 2023-09-21
update sealed-secrets to 0.22.0

yokaze:
Update procedure:
https://github.com/cybozu/neco-containers/blob/main/maintenance.md#sealed-secrets

Jumping from v0.19.5 to v0.24.0 looks a little bit high, so I would like to transit through v0.22.0.


terassyi (approved):
LGTM

PR #1099, opened 2023-10-02
Add mutating webhook to add ipAllowPolicy by annotating the HTTPProxy resource

kmdkuk:
Mutate in .spec.routes[].ipAllowPolicy by annotating the HTTPProxy resource(admission.cybozu.com/ip-policy annotation)
The ipAllowPolicy to be added is managed in the config file.

The following is an example.

config.yaml
```yaml
HttpProxyMutator:
  policies:
  - name: <policy name>
    ipAllowPolicy:
    - source: Peer
      cidr: xxx.yyy.zzz.111
```

httpproxy.yaml
```yaml
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: hoge
  namespace: fuga
  annotations:
    admission.cybozu.com/ip-policy: <policy name>
spec:
  ~~~
  routes:
  - ~~~
    # insert from mutating webhook
    ipAllowPolicy:
    - source: Peer
      cidr: xxx.yyy.zzz.111
```

ref: internal request from NECOREQ-361


admission/hooks/mutate_httpproxy.go
```diff
@@ -64,6 +65,97 @@ func (m *contourHTTPProxyMutator) Handle(ctx context.Context, req admission.Requ
...
+			if err != nil {
+				return err
+			}
+			ipPolicy := map[string]interface{}{}
+			if err := json.Unmarshal(jsonStr, &ipPolicy); err != nil {
+				return err
+			}
+			uniqIpPolicies = append(uniqIpPolicies, ipPolicy)
+		}
+		err = unstructured.SetNestedSlice(route.(map[string]interface{}), uniqIpPolicies, "ipAllowPolicy")
```
kmdkuk, line 123: If I write `uniqIpPolicies = append(uniqIpPolicies, k)` here,
`unstructured.SetNestedSlice` messages `cannot deep copy hooks.HTTPProxyIPFilterPolicy` and panic occurs.
Panic is avoided by converting to `map[string]interface{}` via json.

kmdkuk:
> The order of `[]IPFilterPolicy` in `ipAllowPolicy` does not seem preserved and seems different between multiple create/update (because they are stored in map keys during mutation). This behavior should be noted in the documentation.

Thank you.
Added documantation.
https://github.com/cybozu/neco-containers/pull/1099/commits/65120b451340c41424513eb2724fff2d47c69523

umezawatakeshi (approved):

PR #1123, opened 2023-10-20
promtail-debug: output partial lines to /tmp

umezawatakeshi:
most files in /promtail-debug are copied from /promtail


zoetrope (approved):
LGTM

PR #1138, opened 2023-11-02
Kubernetes v1.27.7

masa213f:
part of https://github.com/cybozu-go/neco/issues/2383

The patch #89155 is no longer needed.
#89155 aimed to avoid [the early return in `syncPod()`](https://github.com/kubernetes/kubernetes/blob/v1.26.10/pkg/kubelet/status/status_manager.go#L682-L685).
But from k8s 1.27, the early return had been removed by https://github.com/kubernetes/kubernetes/pull/115331.


morimoto-cybozu (approved):
If the issue resolved by `89155.patch` still exists in the current Kubernetes, it would be nice to announce the possible undesirable behavior.

masa213f:
@morimoto-cybozu 
The issue will not occur in k8s 1.27. 
The `syncPod()` will be run under the conditions that have caused problems in the past.

PR #1140, opened 2023-11-08
[DNM] ceph-extra-exporter: breakaway from cybozu-go/log and cybozu-go/well

cupnes:
Signed-off-by: Yuma Ogami <yuma-ogami@cybozu.co.jp>


ushitora-anqou:
We merged https://github.com/cybozu/neco-containers/pull/1186, so the role of this PR was over.

PR #1143, opened 2023-11-09
DNM: confirm whether we can bypass ceph-csi issue 2426

satoru-takeuchi:


satoru-takeuchi:
It's just for testing. Close this the test was finished.

PR #1155, opened 2023-11-17
Implement bpf-map-pressure-exporter

tkna:
Signed-off-by: naoki-take <naoki-take@cybozu.co.jp>


tkna:
なんとなく
https://github.com/cybozu/neco-containers/tree/main/ceph-extra-exporter
を参考にしました。

tkna:
https://bozuman.cybozu.com/k/#/space/3801/thread/68598/4795909/9204941
完全なmap名は取れないので、`mapNames`で指定した文字列が含まれているmapについてpressureを取得するようにしました。

bpf-map-pressure-exporter/src/fetcher.go
```diff
@@ -0,0 +1,82 @@
...
+				for itr.Next(&key, &value) {
+					cnt++
+				}
+				results = append(results, bpfMapPressure{
+					mapId:       uint32(id),
+					mapName:     info.Name,
+					mapPressure: float64(cnt) / float64(info.MaxEntries),
+				})
+				break
+			}
```
chez-shanpu, line 78: I think this block can be done concurrently (but it may cause a problem if the number of processes is too large, so we should control it with WaitGroup).

bpf-map-pressure-exporter/src/collector.go
```diff
@@ -0,0 +1,40 @@
...
+	results := c.fetcher.Fetch()
+	for _, result := range results {
+		ch <- prometheus.MustNewConstMetric(
+			c.describe,
+			prometheus.GaugeValue,
+			result.mapPressure,
+			strconv.FormatUint(uint64(result.mapId), 10), result.mapName,
+		)
+	}
+}
```
chez-shanpu, line 40: I think pressure calculation should be done asynchronously and the results are stored somewhere like a global variable, and `Collect()` only fetch map-pressure from that variable when it's called.
Doing calculation in the `Collect()` may cause a problem, because exporter's pressure is too high when the interval for metric collection is shortened and metrics collector gets timeout when the number of target bpf maps are so large.

bpf-map-pressure-exporter/src/fetcher.go
```diff
@@ -0,0 +1,82 @@
...
+		id, err = ebpf.MapGetNextID(id)
+		if errors.Is(err, os.ErrNotExist) {
+			break
+		}
+		if err != nil {
+			_ = logger.Warn("failed to get next map id", map[string]interface{}{
+				"id":        id,
+				log.FnError: err,
+			})
+			return nil
```
chez-shanpu, line 46: How about returning `results` because the results already calculated should not be a problem.

chez-shanpu (commented):
Nice work!
I left some comments.

bpf-map-pressure-exporter/src/fetcher.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			_ = logger.Warn("failed to get map info", map[string]interface{}{
+				"id":        id,
+				log.FnError: err,
+			})
+			return nil
+		}
+		for _, str := range f.mapNameStrings {
+			if strings.Contains(info.Name, str) {
+				itr := m.Iterate()
```
terassyi, line 66: How about using `BatchLookup`.
I'm not sure we can use this for our use case.
We can reduce the number of calling bpf() syscall using this to count elements.
Because `Iterate().Next()` calls bpf() syscall each time...

Please see for details.
- https://pkg.go.dev/github.com/cilium/ebpf#Map.BatchLookup

bpf-map-pressure-exporter/src/fetcher.go
tkna: The process of dumping and counting doesn't need to be concurrent because one execution per map is enough.
It seems inefficient that we use `strings.Contains()` every time in the for loop, but I think it's OK in this case because it's a temporary implementation and the number of `mapNames` is not quite large.

bpf-map-pressure-exporter/src/collector.go
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/7f125beba4c58a82c05741e9bb42ad460474acac

bpf-map-pressure-exporter/src/fetcher.go
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/7f125beba4c58a82c05741e9bb42ad460474acac

bpf-map-pressure-exporter/src/go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/cybozu/neco-containers/bpf-map-pressure-exporter
+
+go 1.21.3
```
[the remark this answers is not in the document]
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/7f125beba4c58a82c05741e9bb42ad460474acac

bpf-map-pressure-exporter/src/main.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := server.ListenAndServe(); err != nil {
+		_ = logger.Critical("failed to ListenAndServe", map[string]interface{}{log.FnError: err})
+		return err
+	}
+	return nil
+}
+
+func main() {
+	port := flag.Uint("port", 8080, "port number")
+	configPath := flag.String("config", "", "config file path")
```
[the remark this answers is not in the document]
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/7f125beba4c58a82c05741e9bb42ad460474acac

bpf-map-pressure-exporter/src/fetcher.go
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/f1ac610f8b7f790f8b4143b8240aa79311fb35d5

bpf-map-pressure-exporter/src/fetcher.go
```diff
@@ -0,0 +1,181 @@
...
+		f.metrics = metrics
+	}()
+
+	metrics = f.fetch()
+}
+
+func (f *bpfMapPressureFetcher) GetMetrics() []bpfMapPressure {
+	f.mutex.RLock()
+	defer f.mutex.RUnlock()
+	return f.metrics
```
chez-shanpu, line 72:
```suggestion
	res := make([]bpfMapPressure, len(f.metrics))
	copy(res, f.metrics)
	return res
```

`f.metrics` has pointers to its elements, so we have to copy it.

bpf-map-pressure-exporter/src/main.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+func startServer(fetcher IBpfMapPressureFetcher, port uint) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	collector := newCollector(fetcher)
+	prometheus.MustRegister(collector)
+	fetcher.Start(ctx)
```
chez-shanpu, line 21:
```suggestion
	go fetcher.Start(ctx)
```

bpf-map-pressure-exporter/src/fetcher.go
```diff
@@ -0,0 +1,181 @@
...
+func (f *bpfMapPressureFetcher) update() {
+	metrics := []bpfMapPressure{}
+
+	defer func() {
+		f.mutex.Lock()
+		defer f.mutex.Unlock()
+		f.metrics = metrics
+	}()
+
+	metrics = f.fetch()
```
chez-shanpu, line 66:
```suggestion
	metrics = f.fetch()
	f.mutex.Lock()
	defer f.mutex.Unlock()
	f.metrics = metrics
```

nit: I'd write like these :)

bpf-map-pressure-exporter/TAG
```diff
@@ -0,0 +1 @@
+0.1.0
```
chez-shanpu, line 1: I think it's good as `1.0.0`

bpf-map-pressure-exporter/src/fetcher.go
```diff
@@ -0,0 +1,181 @@
...
+		info, err := m.Info()
+		if err != nil {
+			_ = logger.Warn("failed to get map info", map[string]interface{}{
+				"id":        id,
+				log.FnError: err,
+			})
+			return results
+		}
+		for _, str := range f.mapNameStrings {
+			if strings.Contains(info.Name, str) {
```
chez-shanpu, line 122:
```suggestion
			if !strings.Contains(info.Name, str) {
				continue
			}
			...
```

nit: I'd write like these :)

terassyi:
There isn't the configuration to build a container.
We want to build a container image at `ghcr.io`.

Please add it like below.
https://github.com/cybozu/neco-containers/blob/34f5cc3beafb874b2be4b51decdba0f7abcdb2e8/.github/workflows/main.yaml#L12-L56C25

bpf-map-pressure-exporter/src/fetcher.go
chez-shanpu: @tkna 

> It seems inefficient that we use strings.Contains() every time in the for loop,

I don't think so. Is there any way to avoid using `strings.Contains()`?

> I think it's OK in this case because it's a temporary implementation and the number of mapNames is not quite large.

I think we should process them in parallel if there is no problem.
However, yes, as you say, we can skip it until we have to do it.

bpf-map-pressure-exporter/src/fetcher.go
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/13b4764764a5c0476d085cdb92880a1432d7d924

bpf-map-pressure-exporter/src/fetcher.go
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/13b4764764a5c0476d085cdb92880a1432d7d924

bpf-map-pressure-exporter/TAG
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/00bcfdd38f8e727229474029a4270aa03d340dc2

bpf-map-pressure-exporter/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# bpf-map-pressure-exporter container
+
+# Stage1: build from source
+FROM quay.io/cybozu/golang:1.21-jammy AS build
+COPY src /work/src
```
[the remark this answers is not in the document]
tkna: https://github.com/cybozu/neco-containers/pull/1155/commits/d1a5fd6d89fdd3226d2d31dce8b4998096ed34cb

terassyi (approved):
LGTM

chez-shanpu (approved):
LGTM

PR #1165, opened 2023-11-22
unbound: add a root trust anchor for DNSSEC validation

yokaze:


masa213f (approved):
LGTM

PR #1186, opened 2023-12-25
ceph-extra-exporter: use log/slog instead of cybozu-go/log

ushitora-anqou:
https://bozuman.cybozu.com/k/59121/show#record=1147

Signed-off-by: Ryotaro Banno <ryotaro.banno@gmail.com>


ceph-extra-exporter/log.go
```diff
@@ -1,10 +1,32 @@
...
+	"os"
+)
 
-var logger *log.Logger
+var logger *slog.Logger
 
 func init() {
-	logger = log.NewLogger()
-	logger.SetFormatter(log.JSONFormat{})
+	utsname, err := os.Hostname()
```
peng225, line 11: Because you get the value from `os.Hostname()` function, I think the variable name should be `hostname` (though `utsname` has almost the same meaning.)

ceph-extra-exporter/log.go
```diff
@@ -1,10 +1,32 @@
...
+				a.Key = "logged_at"
+			case slog.LevelKey:
+				a.Key = "severity"
+			case slog.MessageKey:
+				a.Key = "message"
+			}
+			return a
+		},
+	})).With(
+		slog.String("topic", "ceph-extra-exporter"),
```
peng225, line 29: I think this setting is redundant because there are no other topics in this code.
Because I don't know much about slog package, please let me know if this is some kind of the best practices.

ceph-extra-exporter/executer.go
```diff
@@ -95,29 +94,22 @@ func (ce *cephExecuter) update() {
...
 			continue
 		}
 
 		mv := []metricValue{}
 		if err := json.Unmarshal(result, &mv); err != nil {
-			_ = logger.Warn("parse value failed", map[string]interface{}{
-				"value": string(result),
-				"error": err,
-			})
+			logger.Warn("parse value failed", "value", result, "error", err)
```
peng225, line 112: Could you tell me why you changed `string(result)` to `result`?

IMO, I prefer `string(result)` because `result` comes from the output of the `jq` command, so we can expect that it can be converted to the string type variable.

ceph-extra-exporter/log.go
ushitora-anqou: That's right. I'll fix it.

ceph-extra-exporter/log.go
ushitora-anqou: That's right. I'll fix it.

ceph-extra-exporter/executer.go
ushitora-anqou: I mistakenly thought that `result` was already a string. I'll fix it to use `string(result)`.

ceph-extra-exporter/log.go
```diff
@@ -1,10 +1,29 @@
...
+			case slog.TimeKey:
+				a.Key = "logged_at"
+			case slog.LevelKey:
+				a.Key = "severity"
+			case slog.MessageKey:
+				a.Key = "message"
+			}
+			return a
+		},
+	})).With(slog.String("utsname", hostname))
```
peng225, line 28:
```suggestion
	})).With(slog.String("hostname", hostname))
```

peng225: `utsname` is commonly used, so we do not change it.

peng225 (approved):

PR #1199, opened 2024-01-12
bug fix of squid-exporter

YZ775:
There are some cases that squid-exporter outputs no metrics when multiple request coms at the same time because we call `UnregisterAllMetrics`
This PR fixes this bug.


zeroalphat (approved):
LGTM

PR #1202, opened 2024-01-17
[DNM/CSA] ceph: update to v18.2.1.2

peng225:


peng225:
We found several problems of Ceph v18.2.1, so we skip this version.

PR #1203, opened 2024-01-18
Fix livenessprobe container in envoy

chez-shanpu:
- Rename livenessprobe to probe
- Add `/healthz` and `/readyz` endpoints
- Fix log messages


envoy/pkg/probe/cmd/root.go
```diff
@@ -74,7 +184,12 @@ var rootCmd = &cobra.Command{
 			httpsAddr: config.httpsAddr,
 			timeout:   config.timeout,
 		}
+		lm := &livenessMonitor{m}
+		rm := &readinessMonitor{m}
+
 		mux.Handle("/", m)
+		mux.Handle("/healthz", lm)
+		mux.Handle("/ready", rm)
```
masa213f, line 192: Is there a reason for the presence or absence of `z`?
It would be better to unify the naming convention.

K8s components seems to using the `readyz` as the health-check endpont.
https://kubernetes.io/docs/reference/using-api/health-checks/

masa213f (changes requested):
Please update the README.md too :)

https://github.com/cybozu/neco-containers/blob/bad075ce48cfc27345e1eca75e48a7dd0c01e1bb/envoy/README.md?plain=1#L20

envoy/pkg/probe/cmd/root.go
chez-shanpu: I've fixed :) (And also REAME.md)

masa213f (approved):
LGTM

PR #1205, opened 2024-01-23
Fix unbound image to have prerequisite libs for unbound binaries

yokaze:


chez-shanpu (approved):
LGTM

PR #1209, opened 2024-01-25
Rename scan flag name

yokaze:
### Change Detail

Currently, there are two distinct flags having the same name (`scan`).
One is the input of the `extract_tags` stage, and the other is the output of that.
https://github.com/cybozu/neco-containers/blob/main/.github/actions/extract_tags/action.yaml

This PR changes the input-side name to `request-scan`, while keeping the output-side `scan`.

### Background: BTW, why two distinct scan flags are needed?

Basically, scan is needed when:
- the image is built,
- the image is uploaded, and
- the image is based on a non-scratch image.

This can be computed automatically ... except for some irregular cases.
Normally this happens when base image is substituted by a variable `$X`.
https://github.com/cybozu/neco-containers/blob/main/hubble-relay/Dockerfile#L28

In such cases, CI asks us to set scan requirement manually.
The flag name is currently named `scan`, but it is better to give it another name (`request-scan`).

### DIFF
--- a/.github/actions/build_push/action.yaml
+++ b/.github/actions/build_push/action.yaml
@@ -23,8 +23,8 @@ inputs:
     description: "Run make test before building docker image"
     required: false
     default: "false"
-  scan:
-    description: "Enable image scanning (true, false, or auto)"
+  request-scan:
+    description: "Request image scanning (true, false, or auto)"
     required: false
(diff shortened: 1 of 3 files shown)


chez-shanpu (approved):
LGTM

PR #1210, opened 2024-01-26
Rename extract_tags to prepare_build_params

yokaze:
CI result example for dex:
https://github.com/cybozu/neco-containers/actions/runs/7666357396/job/20893998230


dex/TAG
```diff
@@ -1 +1 @@
-2.37.0.3
+2.37.0.4
```
chez-shanpu, line 1: I think dex's TAG doesn't have to be updated.

chez-shanpu: Ah, I see. You updated it to confirm the workflow working correctly.

chez-shanpu: We've confirmed the actions working correctly, so now, the TAG should be reverted :)

PR #1214, opened 2024-02-02
Migrate local pv provisioner to ghcrio

toshipp:
The PR intends to migrate local-pv-provisioner to ghcr.io. To build it, we first run make check-generate and make test. We had a function for test but not for check-generate, therefore I introduced it. Thanks to it, neco-admission can be built in the build_container job instead of using the dedicated one.

issue: https://github.com/cybozu-private/csa-apps/issues/2487


toshipp:
neco-admission is built successfully.
https://github.com/cybozu/neco-containers/actions/runs/7752293782/job/21141548339?pr=1214

daichimukai (approved):
LGTM

admission/Makefile
```diff
@@ -45,7 +36,7 @@ test: crds simple-test setup-envtest
 	}
 
 .PHONY: simple-test
-simple-test: $(STATICCHECK)
+simple-test: staticcheck
 	test -z "$$(gofmt -s -l . | tee /dev/stderr)"
 	$(STATICCHECK) ./...
```
yokaze, line 41: It should be replaced together?
```suggestion
	staticcheck ./...
```

toshipp: No.

The dependency `staticcheck` is a PHONY target.
https://github.com/cybozu/neco-containers/pull/1214/files#diff-df732126041c7620e3b28705a84144e9aa2f82e739b25f8f1d9f5e452be20fbfR65

On the other hand, the variable refers to the binary.

yokaze: Sorry I missed the fowllowing line and was wondering about how it resolves to a real value.
```
STATICCHECK = $(PROJECT_DIR)/bin/staticcheck
```

yokaze (approved):
LGTM

PR #1215, opened 2024-02-02
Migrate rook and sidecars to ghcrio

toshipp:
issue: https://github.com/cybozu-private/csa-apps/issues/2487


llamerada-jp (commented):
@toshipp 
since entries exist on `.circleci/config.yaml` also, it appears to push images to both quay and ghcr. Is it OK?

toshipp:
@llamerada-jp 
Yes.
I'll stop pushing images after I have the confidence that everything is perfectly fine. But now, I use both.

llamerada-jp (approved):

PR #1221, opened 2024-02-08
Migrate some containers to ghcr.io

yokaze:


go-ipfs/Dockerfile
```diff
@@ -97,4 +98,3 @@ ENV IPFS_LOGGING ""
 ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/start_ipfs"]
 
 CMD ["daemon", "--migrate=true", "--agent-version-suffix=docker"]
-
```
chez-shanpu, line 100: nit: I don't think this line should be removed (files should be end with new line).

.github/workflows/main.yaml
```diff
@@ -87,6 +87,8 @@ jobs:
             container-image: "golang"
             enable_arm64: true
             request-scan: "true"
+          - dir: "./grafana"
+            container-image: "grafana"
```
chez-shanpu, line 91: Probably, grafana should be scanned because the base image is ubuntu (not scratch).

.github/workflows/main.yaml
```diff
@@ -89,6 +89,8 @@ jobs:
             request-scan: "true"
           - dir: "./grafana"
             container-image: "grafana"
+          - dir: "./grafana_plugins_init"
+            container-image: "grafana_plugins_init"
```
chez-shanpu, line 93: Also, I think grafana_plugins_init should be scanned as grafana is.

.github/workflows/main.yaml
yokaze: Currently scan requirement is auto-detected, so I think there would be no problem.
ref: https://github.com/cybozu/neco-containers/actions/runs/7828640467/job/21358962831#step:4:178

.github/workflows/main.yaml
yokaze: Same as the previous one:
ref: https://github.com/cybozu/neco-containers/actions/runs/7828640467/job/21358963112#step:4:178

go-ipfs/Dockerfile
yokaze: There were two LFs at the end of the file, so I removed one.
```
$ git checkout main
Already on 'main'
Your branch is up to date with 'origin/main'.
$ hexdump -C go-ipfs/Dockerfile | tail -n 3
000009e0  6e 2d 73 75 66 66 69 78  3d 64 6f 63 6b 65 72 22  |n-suffix=docker"|
000009f0  5d 0a 0a                                          |]..|
000009f3
```

.github/workflows/main.yaml
chez-shanpu: I see, it looks good as is :)

chez-shanpu (approved):
LGTM

PR #1223, opened 2024-02-13
Update envoy and contour

chez-shanpu:


chez-shanpu:
depends on https://github.com/cybozu/neco-containers/pull/1224

envoy/go.mod
```diff
@@ -1,33 +1,39 @@
 module github.com/cybozu/neco-containers/envoy
 
-go 1.20
+go 1.21
+
+toolchain go1.21.4
```
YZ775, line 5: Is this directive required?

YZ775:
I think we need to update go version in envoy's Dockerfile

https://github.com/cybozu/neco-containers/blob/b664e5ccd3283f0f89574dcd863b183a9b100674/envoy/Dockerfile#L4

envoy/go.mod
chez-shanpu: It wouldn't be required. I think it was added somehow when I update its dependencies. I'll remove it :)

YZ775 (approved):
LGTM

PR #1224, opened 2024-02-13
Revert "Extract common tool setup" in admission

chez-shanpu:
This PR reverts changes about admission from "Extract common tool setup"(e9c70f89) to fix docker build error.


YZ775 (approved):
LGTM

PR #1227, opened 2024-02-14
Add generic make-targets step

yokaze:
This PR:
- adds `make-targets` parameter to list all the required make-targets before build and
- moves most of the test steps to GitHub Actions (except envoy).


admission/TAG
```diff
@@ -1 +1 @@
-0.24.5
+0.24.6
```
morimoto-cybozu, line 1: This conflicts with https://github.com/cybozu/neco-containers/pull/1220.

```suggestion
0.24.7
```

morimoto-cybozu:
This comment is not a change request.

I prefer a set of small commits to a large commit for complicated changes.
It's hard for me to check corresponding addition and deletion for each software components.

morimoto-cybozu (changes requested):

PR #1232, opened 2024-02-16
Fix implementation of parsing service_times in squid_exporter

zeroalphat:
The `service_times` metric produced by squid seems to change the number of spaces used as delimiters as the number of digits in the value increases.
The current implementation of the squid exporter parses the value based on the assumption that there are two spaces, so if there is a single space, the value will not be parsed correctly.

```
HTTP Requests (All):  90%   0.00123  0.00345
HTTP Requests (All):  95%   0.01234 80.12345
```

This PR changes the implementation of squid_exporter to allow parsing of variable whitespace used as delimiters.


YZ775 (approved):
I didn't know that.
Thank you!

PR #1233, opened 2024-02-29
ceph: build the previous version

peng225:
I changed the structure under `ceph` directory so that we can build multiple versions of Ceph. I also changed the CI jobs (both CircleCI and Github Actions) for this purpose.

Because Ceph v17 build was not migrated to ghcr.io, I did it, too.


toshipp (commented):
I'm considering whether we need to maintain both X.Y.Z.W and X.Y.Z.W'.
If we upgrade ceph from v17 to v18, we first update rook base image to v18. After that, we will try to update ceph cluster to v18, but if it does not work properly, we must use v17. In this case, we must maintain both versions.

On the other hand, if we update ceph from 17.2.1.1 to 17.2.1.2, and rook works properly but ceph cluster does not, we will make a new ceph image 17.2.1.3. In this case, we use it for rook and ceph, so we need only one image.

I think it is sufficient to maintain one directory per upstream version X.Y.Z. What do you think?

peng225:
> I think it is sufficient to maintain one directory per upstream version X.Y.Z. What do you think?

I agree with it. I changed the code as you proposed.

maintenance.md
```diff
@@ -164,14 +164,39 @@ In Regular update, do the following as part of the update of each CRD-providing
...
    2. If there are ceph runtime packages or required tool changes, update Dockerfile.
-3. Update the `version` argument on the `build-ceph` job in the CircleCI `main` workflow and the `build_ceph` job in the Github Actions `main` workflow.
-4. Update `BRANCH` and `TAG` files.
+4. Update the `version` argument on the `build-ceph-latest` job in the CircleCI `main` workflow and the `build_ceph` job in the Github Actions `main` workflow.
+5. Update `BRANCH` and `TAG` files in the `ceph/latest` directory.
 
-***NOTE:*** The rook image is based on the ceph image. So upgrade the [rook](#rook) image next.
+### When you update based on the `cephX-Y-Z` directory
+
+1. If the **new** version is vA.B.C.D, then copy the `ceph/vX-Y-Z` directory to the `ceph/vA-B-C` directory and add the following CI jobs.
```
toshipp, line 192: vX-Y-Z is the upstream version. If we need to change files on the directory, it is our artifacts (Dockerfile, build scripts, or patches). I think we don't need to copy this because the upstream version has not changed.
Could you explain your thoughts with specific version numbers?

toshipp: Got it.
I prefer waiting for a new v18 version, but it would be better to discuss this with the whole team.

peng225:
@toshipp Beause we have decided to drastically change the way to build the previous versions of Ceph, I'll close this PR and create a new one.

PR #1236, opened 2024-03-07
Add tcp-keepalive server and client container image

terassyi:
This PR add the `tcp-keepalive` container that performs TCP server and client to exchange keepalive messages each other.

This container is for checking the long live connection.


chez-shanpu (approved):
LGTM

PR #1277, opened 2024-03-14
Rebuild s3gw with go-lang 1.22 and dependencys

takara9:
This PR is the part of the regular container update.
Some module upgradeed and code adjusted for module changes.


s3gw/go.mod
```diff
@@ -1,40 +1,39 @@
 module github.com/cybozu/neco-containers/s3gw
 
-go 1.20
+go 1.22.1
```
umezawatakeshi, line 3: Patch versions should not be set.

umezawatakeshi (changes requested):

s3gw/go.mod
takara9: After installing Go 1.21, we never have to manually download and install a Go toolchain again. 
The go command can do it for you. but, we must specify such as 1.21.0.

Please refer this link for detail 
https://go.dev/blog/toolchain

umezawatakeshi (approved):

umezawatakeshi:
I approved but test did not passed.

PR #1286, opened 2024-03-22
add ttypdb: create pdbs on the bases of ttys in pods.

umezawatakeshi:


ttypdb/src/common/http.go
```diff
@@ -0,0 +1,58 @@
...
+		orig:   orig,
+		logger: logger,
+	}
+}
+
+func (h *proxyHTTPHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	start := time.Now()
+
+	prw := &proxyHTTPResponseWriter{
+		orig: rw,
```
morimoto-cybozu, line 31:
```suggestion
		ResponseWriter: rw,
```

ttypdb/src/common/http.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"net/http"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+type proxyHTTPHandler struct {
+	orig   http.Handler
```
morimoto-cybozu, line 11:
```suggestion
	http.Handler
```

ttypdb/src/controller/poll.go
```diff
@@ -0,0 +1,177 @@
...
+		if foundPdb {
+			err := pdbInterface.Delete(ctx, pod.Name, metav1.DeleteOptions{})
+			if err != nil {
+				logger.Error("failed to delete PDB", zap.Error(err))
+			} else {
+				logger.Info("deleted PDB")
+			}
+		} else {
+			logger.Debug("PDB does not exist")
+		}
```
morimoto-cybozu, line 142: You might use the "return early" pattern.

ttypdb/src/sidecar/count.go
```diff
@@ -0,0 +1,59 @@
...
+
+			statBytes, err := os.ReadFile(filepath.Join("/proc", name, "stat"))
+			if err != nil {
+				return err
+			}
+
+			// The 6th (0-origin) field is controlling tty device number.
+			// If it is "0", the process is not controlled.
+			stat := string(statBytes)
+			idx := strings.LastIndexByte(stat, ')')
```
morimoto-cybozu, line 38: It would be simpler if we could use `fmt.Sscanf()`.
Do we need to care about a command name with spaces?

umezawatakeshi: Yes, wee need to do so.
procps (`ps` command) do the same thing: https://gitlab.com/procps-ng/procps/-/blob/v4.0.4/library/readproc.c?ref_type=tags#L578-581

umezawatakeshi:
I also added a care for implicit status in proxyHTTPResponseWriter.

morimoto-cybozu (approved):
LGTM

PR #1289, opened 2024-03-26
unbound: embed root.hints in container

yokaze:
This PR embeds `root.hints` file in unbound container and makes necessary improvements to common infrastructure:

1. adds post-build step
  I want to include in the CI output the content of the embedded `root.hints`, so I added `make-post-targets` flag that are executed after building an image.
2. adds `load: true` flag to `docker/build-push-action`
  This flag allows us to use the image in GitHub Actions.
  See: https://docs.docker.com/build/ci/github-actions/test-before-push/


.github/actions/build_push/action.yaml
```diff
@@ -69,13 +73,24 @@ runs:
...
+    - name: Make post targets
+      if: ${{ inputs.make-post-targets != '' && steps.prepare.outputs.build }}
+      shell: bash
+      env:
+        IMAGE_TAG: ${{ steps.prepare.outputs.tag }}
+      run: |
+        for i in ${{ inputs.make-post-targets }}; do
+          echo "Executing make $i..."
+          make -C ${{ inputs.dir }} $i
+        done
```
chez-shanpu, line 93: I think the docker image should be pushed after `make`.

chez-shanpu: Also, maybe we should scan the image before pushing it.

chez-shanpu (commented):
I left several comments :)

chez-shanpu (approved):
LGTM!

PR #1292, opened 2024-03-29
delete grafana_plugins_init, which is no longer used by grafana-operator

umezawatakeshi:
part of https://github.com/cybozu-private/neco-tasks/issues/57

Grafana operator v5 does not use grafana_plugins_init.


morimoto-cybozu (approved):
LGTM

PR #1295, opened 2024-04-01
Fix test envoy

tkna:
Add contition to "test envoy".

https://github.com/cybozu/neco-containers/actions/runs/8505518042/job/23294131047


yokaze (approved):
LGTM 🙇🙇

PR #1298, opened 2024-04-04
Update golang to 1.22.2 and 1.21.9

chez-shanpu:
Update golang versions.
Also, this PR makes docker_push action does not enable the load option when the image is built with multi-platform.


YZ775 (approved):
LGTM

PR #1300, opened 2024-04-04
Update placemat in meows-dctest-runner image

masa213f:
placemat v2.4.5 is released. So I update the meows-dctest-runner.
ref: https://github.com/cybozu-go/placemat/releases/tag/v2.4.5


zoetrope (approved):
LGTM

PR #1333, opened 2024-04-22
Update envoy to 1.29.4 and its dependencies

chez-shanpu:


envoy/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cybozu/neco-containers/envoy
 
-go 1.21
+go 1.22
```
masa213f, line 3: I guess we need specify the patch version.

```suggestion
go 1.22.2
```

masa213f (changes requested):

envoy/go.mod
chez-shanpu: @masa213f  Wny do you think so?
My understandig is that it is treated equally with `x.xx.0` when we don't specify the patch.

masa213f: I faced a build error a while ago without the "go toolchain" directive.
And I commented that I wondered if it would be the same this time.
If not, that's fine.

chez-shanpu: I googled, and it seems that we should specify the patch version after `1.21.x`.
https://github.com/golang/go/issues/62278

I think this can be built with 1.22.x so, I'll specify the version `1.22.0`.

masa213f (approved):

PR #1343, opened 2024-05-08
kube-state-metrics and kube-metrics-adapter supports

terassyi:


yokaze (approved):
LGTM

PR #1345, opened 2024-05-08
moco-switchover-downtime-monitor supports k8s 1.28

terassyi:


yokaze (approved):
moco-switchover-downtime-monitor is marked to be updated within a regular update, possibly no need to update within k8s update?

https://github.com/cybozu/neco-containers/blob/main/maintenance.md#moco-switchover-downtime-monitor

terassyi:
That's certainly true.
But I think it's ok to update at this time.

PR #1350, opened 2024-05-29
container update to support k8s 1.29

umezawatakeshi:
part of https://github.com/cybozu-private/neco-tasks/issues/330

coredns, etcd, kube-metrics-adapter, kube-state-metrics, pause and unbound_exporter are already latest.


admission/Makefile
```diff
@@ -1,9 +1,9 @@
 CONTROLLER_TOOLS_VERSION = 0.15.0
-KUSTOMIZE_VERSION = 5.4.1
+KUSTOMIZE_VERSION = 5.4.2
 CONTOUR_VERSION = 1.28.2
 ARGOCD_VERSION = 2.10.7
 GRAFANA_OPERATOR_VERSION = 5.8.1
-ENVTEST_K8S_VERSION = 1.28.3
+ENVTEST_K8S_VERSION = 1.29.3
```
YZ775, line 6: I think the latest envtest is 1.29.4, shouldn't we use that?

pod-delete-rate-limiter/Makefile
```diff
@@ -2,7 +2,7 @@
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.28.3
+ENVTEST_K8S_VERSION = 1.29.3
```
YZ775, line 5: ditto

admission/Makefile
umezawatakeshi: That's because the envtest installed by this Makefile does not support 1.29.4 . (It depends on the version of controller-runtime)
However, we decided to update all modules and this would be modified accordingly.

umezawatakeshi: I finally found that the latest *released* controller-runtime (v0.18.3) only supports 1.29.3.
(The `controller-runtime@latest` supports 1.29.5)

YZ775 (approved):
LGTM

PR #1360, opened 2024-06-19
Disable libz3

tkna:


.github/actions/build_cilium-image-tools/action.yaml
```diff
@@ -50,10 +50,29 @@ runs:
   - name: Build llvm
     shell: bash
     run: |
+      WORKSPACE=$(pwd)/cilium/src/workspace
       cd cilium/src/image-tools
       sudo images/llvm/checkout-llvm.sh
-      sudo images/llvm/build-llvm-native.sh
-      sudo mv /out/linux/amd64/bin/clang /out/linux/amd64/bin/llc /out/linux/amd64/bin/llvm-objcopy ../workspace/bin
+      
+      # images/llvm/build-llvm-native.sh
```
terassyi, line 57: For the maintainability, I think we need to update `maintenance.md`.

But I'm not sure that we should update it in this PR.
It may be ok after the procedure is finalized.

tkna: > It may be ok after the procedure is finalized.

I think this is better.

terassyi (approved):
LGTM

PR #1373, opened 2024-07-08
Add tcp-keepalive metrics

chez-shanpu:
This PR adds the metrics below.

- `connection`
  - A TCP connection status
- `receive_total`
  - num of receive messages
- `send_total`
  - num of send messages
- `retry_count`
  - num of consecutive retries
- `retry_total`
  - num of total retries

### DIFF
--- a/tcp-keepalive/internal/server/server.go
+++ b/tcp-keepalive/internal/server/server.go
@@ -8,12 +8,15 @@ import (
 	"log/slog"
 	"net"
 	"os"
+
+	"github.com/neco-containers/tcp-keepalive/internal/metrics"
 )
 
 var log *slog.Logger
 
 type Server struct {
 	listener *net.TCPListener
+	metrics  *Metrics
 
 	*Config
 }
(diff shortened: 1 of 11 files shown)


tcp-keepalive/internal/server/server.go
```diff
@@ -26,18 +29,31 @@ func initLogger() {
 	log = slog.New(slog.NewJSONHandler(os.Stdout, nil))
 }
 
-func NewServer(cfg *Config) (*Server, error) {
+func NewServer(cfg *Config, mcfg *metrics.Config) (*Server, error) {
```
terassyi, line 32: I think we don't have to receive pointer type as arguments.
If we accept pointer types, we have to check whether they are nil or not.

tcp-keepalive/internal/server/server.go
```diff
@@ -26,18 +29,31 @@ func initLogger() {
...
+	log.Info("new server", slog.Any("config", cfg), slog.Any("metrics", m))
+	return &Server{Config: cfg, metrics: m}, nil
 }
 
 func (s *Server) Run(ctx context.Context) error {
 	log = log.With("listen", s.ListenAddr)
 
+	if s.metrics.Export {
+		go func() {
+			for {
```
terassyi, line 51: Do we need this loop?

tcp-keepalive/internal/server/server.go
```diff
@@ -26,18 +29,31 @@ func initLogger() {
...
+				log.Error("serving metrics failed", slog.Any("error", s.metrics.Serve().Error()))
```
terassyi, line 52: I think it is better to split lines `s.metrics.Serve()` and log output.

tcp-keepalive/internal/server/server.go
chez-shanpu: @terassyi 
> I think we don't have to receive pointer type as arguments.

Yes, we don't have to use pointer types but I think we should use them because they have an advantage in terms of memory footprint compared to value types.

> If we accept pointer types, we have to check whether they are nil or not.

I agree. I'll add nil checks.

tcp-keepalive/internal/server/server.go
chez-shanpu: @terassyi 
Without this loop, tcp-keepalive will run but won't export any metrics when the metrics server fail with an error.

tcp-keepalive/internal/server/server.go
chez-shanpu: @terassyi Why do you think so? Is it for readability?

chez-shanpu: It might not be good to include the server Serve() in a log output line.

chez-shanpu: I split `s.metrics.Serve()` (also, the client too).

terassyi (approved):

PR #1385, opened 2024-07-19
Fix GraphQL queries to the correct syntax

zeroalphat:


umezawatakeshi:
I cannot review it because I don't know about GraphQL syntax.

morimoto-cybozu (approved):

PR #1423, opened 2024-09-12
update kubernetes to 1.30.5

YZ775:


chez-shanpu (commented):
Is it ok that 1.30.4 remains in `pause/Dockerfile` and `pod-delete-rate-limiter/Makefile`?

YZ775:
@chez-shanpu 
Thank you for reviewing,

There are no diff between 1.30.4 and 1.30.5, I think we don't need to update pause.
https://github.com/kubernetes/kubernetes/tree/v1.30.4/build/pause
https://github.com/kubernetes/kubernetes/tree/v1.30.5/build/pause

In addition, `ENVTEST_K8S_VERSION` in `pod-delete-rate-limiter/Makefile` is only used for test. So I think we don't need to update.

chez-shanpu:
@YZ775 I see. As you say,  they don't seem to need updating.
I'll approve it :)

chez-shanpu (approved):
LGTM

YZ775:
Thank you!

PR #1436, opened 2024-09-24
Create update.yaml 

012xx:
This PR updates the Go version to the latest stable version using CI automation.

## Updated files:
- `golang-all/golang-1.22-jammy/Dockerfile`
- `golang-all/golang-1.22-jammy/TAG`


.github/workflows/update.yaml
takahiro-yamada, line 1: It would be better if implicit policies, such as 'We need to maintain Ubuntu versions and Go minor versions manually,' were written as a comment.

.github/workflows/update.yaml
```diff
@@ -0,0 +1,54 @@
...
+        if: env.NEED_UPDATE == 'true'
+        run: |
+          stable_version=${STABLE_GOLANG_VERSIONS[0]}
+          major_minor=$(echo $stable_version | cut -d. -f1,2)
+          major_minor_patch=$(echo $stable_version | cut -d. -f1,2,3)
+
+          # Update golang-all/golang-1.22-jammy/Dockerfile
+          sed -i "s/ARG GO_VERSION=.*/ARG GO_VERSION=${stable_version}/" golang-all/golang-1.22-jammy/Dockerfile
+          # Update golang-all/golang-1.22-jammy/TAG
+          sed -i "s/[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+_jammy/${major_minor_patch}.1_jammy/" golang-all/golang-1.22-jammy/TAG
```
takahiro-yamada, line 43: Values that may vary and are repeatedly referenced such as `1.22` and `jammy` can be made into variables for ease of maintenance.

takahiro-yamada (changes requested):

.github/workflows/update.yaml
```diff
@@ -0,0 +1,54 @@
...
+      GH_TOKEN: ${{ secrets.CYBOZU_NECO_PAT }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Get latest stable Go versions
+        id: get-go-versions
+        run: |
+          # Get latest stable Go versions(x.xx.x)
+          stable_golang_versions=($(curl -sSL https://go.dev/dl/?mode=json | jq -r '[.[] | select(.stable == true)][1] | .version | sub("^go"; "")'))
```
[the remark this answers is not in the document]
012xx: The following intentions are behind this. If there are no problems with the background, I want to leave the commands as they are (the comments will be modified).

# Get the second most stable version of Go (x.xx.x).
# Updating to the latest version requires creating and deleting directories, etc., and it takes time to automate, so it is currently not supported.

PR #1440, opened 2024-10-02
Scan hubble-ui-frontend and hubble-ui-backend

zoetrope:


yokaze (changes requested):
If a base image is substituted from a variable, CI cannot detect if base is scratch or ubuntu. Therefore we need to set `request-scan` field manually.
CI reports it as a failure.

Could you please add `request-scan: "true"`?

yokaze (approved):

PR #1442, opened 2024-10-11
Update GitHub Actions workflow for Go version updates

012xx:
This PR updates the Go version to the latest stable version using CI automation.

Updated files:
- `golang-all/golang-$go_major-$CODENAME/Dockerfile`
- `golang-all/golang-$go_major-$CODENAME/TAG`

e.g.)
- `golang-all/golang-1.23-jammy/Dockerfile`
- `golang-all/golang-1.23-jammy/TAG`
- `golang-all/golang-1.22-jammy/Dockerfile`
- `golang-all/golang-1.22-jammy/TAG`

### DIFF
--- a/.github/workflows/update.yaml
+++ b/.github/workflows/update.yaml
@@ -27,12 +27,15 @@ jobs:
           touch ./update-body.txt
           
           # Get latest stable Go versions
-          GO_STABLE_MAJOR_VERSIONS=$(curl -sSLf https://go.dev/dl/?mode=json | jq -r '[.[] | select(.stable == true) | .version | sub("^go"; "") | split(".")[:2] | join(".")] | unique | .[]')
-          if [ -z "${GO_STABLE_MAJOR_VERSIONS}" ]; then
+          GO_STABLE_VERSIONS=$(curl -sSLf https://go.dev/dl/?mode=json | jq -r '[.[] | select(.stable == true) | .version | sub("^go"; "")] | unique | .[]')
+          if [ -z "${GO_STABLE_VERSIONS}" ]; then
             echo "Failed to fetch stable Go versions"
             exit 1
           fi
 
+          # Extract major and minor versions for comparison
+          GO_STABLE_MAJOR_VERSIONS=$(echo "${GO_STABLE_VERSIONS}" | cut -d. -f1,2 | sort -u)
+          
           # Check for differences between GO_MAJOR_VERSIONS and GO_STABLE_MAJOR_VERSIONS
           DIFF=$(diff <(printf "%s\n" ${GO_MAJOR_VERSIONS} | sort -V) <(printf "%s\n" ${GO_STABLE_MAJOR_VERSIONS} | sort -V))
           if [ -n "${DIFF}" ]; then
@@ -49,14 +52,11 @@ jobs:
             local latest_version=$3
             local dockerfile_path="golang-all/golang-${go_major}-${codename}/Dockerfile"
             local tag_path="golang-all/golang-${go_major}-${codename}/TAG"
-
             if [ ! -f "${dockerfile_path}" ] || [ ! -f "${tag_path}" ]; then
               echo "Files not found for Go ${go_major} on Ubuntu ${codename}"
               return
             fi
-
             local current_version=$(grep 'ARG GO_VERSION=' "${dockerfile_path}" | cut -d= -f2)
-
             if [ "${current_version}" != "${latest_version}" ]; then
               NEED_UPDATE=1
               # Update Dockerfile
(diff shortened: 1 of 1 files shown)


rkwmt (changes requested):

.github/workflows/update.yaml
```diff
@@ -0,0 +1,103 @@
...
+          git config --global user.email "neco@cybozu.com"
+          git config --global user.name "cybozu-neco"
+          
+          BRANCH=update-go-$VERSIONS
+          git checkout -b $BRANCH
+          git add -u
+          git commit -m "Update Go versions to $VERSIONS"
+          git push origin $BRANCH
+          
+          PR_TITLE="Update Go versions to $VERSIONS"
```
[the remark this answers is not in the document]
012xx: It was unintentional. Thank you for pointing this out🙏

.github/workflows/update.yaml
```diff
@@ -0,0 +1,107 @@
...
+              DOCKERFILE_PATH="golang-all/golang-$go_major-$CODENAME/Dockerfile"
+              TAG_PATH="golang-all/golang-$go_major-$CODENAME/TAG"
+              
+              if [ ! -f "$DOCKERFILE_PATH" ] || [ ! -f "$TAG_PATH" ]; then
+                echo "Files not found for Go $go_major on Ubuntu $CODENAME"
+                continue
+              fi
+              
+              CURRENT_VERSION=$(grep 'ARG GO_VERSION=' "$DOCKERFILE_PATH" | cut -d= -f2)
+              CURRENT_TAG=$(cat "$TAG_PATH" | cut -d_ -f1)
```
takahiro-yamada, line 75: `CURRENT_TAG` doesn't seem to be used.

takahiro-yamada (changes requested):

.github/workflows/update.yaml
```diff
@@ -0,0 +1,106 @@
...
+            echo "Failed to fetch stable Go versions"
+            exit 1
+          fi
+
+          # Check if there are newer major versions available
+          HIGHEST_MAJOR_VERSION=$(echo "$STABLE_VERSIONS" | cut -d. -f1-2 | sort -V | tail -n1)
+          echo "HIGHEST_MAJOR_VERSION="$HIGHEST_MAJOR_VERSION
+          HIGHEST_CONFIGURED_VERSION=$(echo "${{ env.GO_MAJOR_VERSIONS }}" | tr ' ' '\n' | sort -V | tail -n1)
+          echo "HIGHEST_CONFIGURED_VERSION="$HIGHEST_CONFIGURED_VERSION
+          if [ "$(echo -e "$HIGHEST_MAJOR_VERSION\n$HIGHEST_CONFIGURED_VERSION" | sort -V | tail -n1)" = "$HIGHEST_MAJOR_VERSION" ] && [ "$HIGHEST_MAJOR_VERSION" != "$HIGHEST_CONFIGURED_VERSION" ]; then
```
rkwmt, line 41: I feel this code for checking newer versions is complex. To make it simpler, how about detecting only the differences between the stable versions of Go and the versions we specify in `GO_MAJOR_VERSIONS` ? For example, we could use the `diff` command and write something like this:

```shell
GO_MAJOR_VERSIONS="1.23 1.22"
GO_STABLE_MAJOR_VERSIONS="1.24 1.23"
diff <(printf "%s\n" ${GO_MAJOR_VERSIONS}) <(printf "%s\n" ${GO_STABLE_MAJOR_VERSIONS})
```

.github/workflows/update.yaml
```diff
@@ -0,0 +1,113 @@
...
+            
+            for go_major in ${GO_MAJOR_VERSIONS}; do
+              # Find matching version from stable versions
+              LATEST_VERSION=$(echo "${GO_STABLE_MAJOR_VERSIONS}" | grep "^${go_major}" | head -n1)
+              if [ -z "${LATEST_VERSION}" ]; then
+                echo "No matching stable version found for Go ${go_major}"
+                continue
+              fi
+              
+              update_files "${go_major}" "${CODENAME}" "${LATEST_VERSION}"
```
rkwmt, line 90: `LATEST_VERSION` does not include a minor revision number, but it is compared to `GO_VERSION` which includes a minor revision number. I think this is unintended behavior, please fix.
https://github.com/cybozu/neco-containers/blob/4b958bc1b3c19f46e70fd80ac4277d9b7c7ffd71/.github/workflows/update.yaml#L58-L60

rkwmt (changes requested):

.github/workflows/update.yaml
```diff
@@ -0,0 +1,112 @@
...
+              echo "Files not found for Go ${go_major} on Ubuntu ${codename}"
+              return
+            fi
+            local current_version=$(grep 'ARG GO_VERSION=' "${dockerfile_path}" | cut -d= -f2)
+            if [ "${current_version}" != "${latest_version}" ]; then
+              NEED_UPDATE=1
+              # Update Dockerfile
+              sed -i "s/ARG GO_VERSION=.*/ARG GO_VERSION=${latest_version}/" "${dockerfile_path}"
+              # Update TAG file
+              echo "${latest_version}_${codename}" > "${tag_path}"
```
rkwmt, line 65: The format of the TAG looks like `X.Y.Z.N-jammy`, where `X.Y.Z` is the Go version and N starts at 1 and increments. Therefore, it seems better to set it like `X.Y.Z.1_jammy` when updating.
https://github.com/cybozu/neco-containers/blob/8ce1748be91631f9cf478c22b17c06748abc04c4/golang-all/golang-1.22-jammy/TAG#L1

rkwmt (changes requested):

takahiro-yamada (approved):
LGTM

PR #1465, opened 2024-11-13
Update kube-storage-version-migrator, registry, and vault

zeroalphat:


vault/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN make bootstrap && \
     make
 
 # Stage2: setup runtime container
-FROM ghcr.io/cybozu/ubuntu:22.04
+FROM ghcr.io/cybozu/ubuntu:24.04
```
morimoto-cybozu, line 17: @zeroalphat 
The runtime container will now use a different version of Ubuntu than that for the build.
If you have confirmed that the binaries are statically built, this would be safe.

zeroalphat: I fixed it so that the Ubuntu versions for build and runtime are the same.

morimoto-cybozu (approved):

PR #1471, opened 2024-11-15
Update admission to 0.92.2

masa213f:
- **Update go modules**
- **Fix URL**


morimoto-cybozu (approved):
This PR looks perfect to me!

BTW, `admission` would be updated in (1) Kubernetes update and (2) regular update of each of (2-a) Contour, (2-b) Argo CD, and (2-c) Grafana Operator according to the current `maintenance.md`.
It would be good to not include `admission` as a separate item in the list of regular update candidates.

PR #1492, opened 2024-12-16
Fix matrix generator to handle deleted files

yokaze:
Previously, CI matrix generator failed when relevant files were to be deleted.
This PR fixes the script for the problem.


terassyi (approved):
LGTM

PR #1496, opened 2024-12-26
Update argocd-image-updater version to v0.15.1

rkwmt:


argocd-image-updater/Dockerfile
```diff
@@ -10,15 +10,15 @@ WORKDIR /work/argocd-image-updater
...
 
 RUN groupadd -g 10000 argocd && \
     useradd -r -u 10000 -g argocd argocd && \
     mkdir -p /home/argocd && \
     chown argocd:argocd /home/argocd && \
     apt-get update && \
-    apt-get install -y --no-install-recommends git openssh-client python3 python3-pip && \
+    apt-get install -y --no-install-recommends git openssh-client python3 python3-pip tini gpg gpg-agent && \
     pip3 install --upgrade pip && \
     pip3 install awscli && \
```
takahiro-yamada, line 23: `24.04/ubuntu-debug` installs awscli with `curl`, how about using this method?
With this, we can omit installing `python3` and `python3-pip` I think.
https://github.com/cybozu/ubuntu-base/blob/1cb03b74e6b001c989ea83296a035e1bb9b7d129/24.04/ubuntu-debug/Dockerfile#L55-L58

argocd-image-updater/Dockerfile
```diff
@@ -10,15 +10,15 @@ WORKDIR /work/argocd-image-updater
...
     apt-get clean && \
```
takahiro-yamada, line 24: We can omit explicit `apt-get clean`.

https://docs.docker.com/build/building/best-practices/
> Official Debian and Ubuntu images [automatically run apt-get clean](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required.

argocd-image-updater/Dockerfile
```diff
@@ -10,15 +10,15 @@ WORKDIR /work/argocd-image-updater
...
+FROM ghcr.io/cybozu/ubuntu:22.04
 LABEL org.opencontainers.image.source="https://github.com/cybozu/neco-containers"
 
 RUN groupadd -g 10000 argocd && \
     useradd -r -u 10000 -g argocd argocd && \
     mkdir -p /home/argocd && \
     chown argocd:argocd /home/argocd && \
     apt-get update && \
-    apt-get install -y --no-install-recommends git openssh-client python3 python3-pip && \
+    apt-get install -y --no-install-recommends git openssh-client python3 python3-pip tini gpg gpg-agent && \
```
takahiro-yamada, line 21: I think it's better to write package names line by line and alphabetize them.

takahiro-yamada (changes requested):

takahiro-yamada (approved):
LGTM

PR #1504, opened 2025-01-17
Update cilium-operator to metallb patched version(v1.14.14)

terassyi:
cilium-operator-generic container is built based on https://github.com/cybozu-go/cilium/tree/v1.14.14-support-metallb-requiring-ip from this PR.

See following for detail changes.
https://github.com/cybozu-go/cilium/commit/357f8dc11b3f2326fbdc51dabe328e3bd84c085e


terassyi:
we don't need this anymore.

PR #1509, opened 2025-01-21
revise the rook and ceph image update procedure

peng225:

### DIFF
--- a/maintenance.md
+++ b/maintenance.md
@@ -274,9 +274,6 @@ TBD
 3. Update the `version` argument on the `build-ceph` job in the CircleCI `main` workflow and the `build_ceph` job in the Github Actions `main` workflow.
 4. Update `BRANCH` and `TAG` files.
 
-> [!Note]
-> The rook image is based on the ceph image. So upgrade the [rook](#rook) image next.
-
 ### Create a patched image from the specific version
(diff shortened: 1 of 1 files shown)


maintenance.md
```diff
@@ -952,12 +946,15 @@ TBD
 - csi-snapshotter
 
 1. Check the [release page](https://github.com/rook/rook/releases).
-2. Check the upstream Dockerfile. If there are any updates, update our `Dockerfile`.
+2. Check the `CEPH_VERSION` of the upstream Makefile. If the version is changed, update the base image version in our `Dockerfile`.
```
llamerada-jp, line 949: I think we used the ceph even if its version is different from the one in the makefile when the version of it is supported.

peng225: In my understanding, it seems that at least the major version of Ceph was implicitly aligned with the upstream Makefile of Rook. My proposal is that we should write the policy in the maintenance document (and should not use the older majour version of Ceph even if that is supported by Rook).

Moreover, I believe we should use the exact same version of Ceph as the one in Rook's upstream Makefile. If we don't, there's no guarantee that those version combinations will work properly.

llamerada-jp (approved):

PR #1522, opened 2025-02-04
Add bpftrace container

yokaze:
This PR:
- adds `ghcr.io/cybozu/bpftrace` container.
- tweaks `build_push` action to accept `platforms` and `load` parameters.
- bumps up golang-1.23-noble for verification.


bpftrace/Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+        libfl-dev \
+        python3-setuptools \
+        liblzma-dev \
+        libdebuginfod-dev \
+        libpolly-18-dev
+
+# Missing -DREVISION causes an error, although it is not documented
+# ref. https://github.com/iovisor/bcc/blob/master/cmake/version.cmake
+RUN mkdir build && \
+    cmake -B build -DREVISION=0.33.0 && \
```
terassyi, line 72: `REVISION` is hard coded here.
I suspect that this is bcc version, right?
If so, it would be better to use ARG or to document somewhere.

yokaze: Sorry I missed it!

terassyi (approved):
LGTM

PR #1534, opened 2025-02-14
Update bmc-reverse-proxy for k8s 1.31

chez-shanpu:


masa213f (commented):
Could you update the golang image?
https://github.com/cybozu/neco-containers/blob/main/bmc-reverse-proxy/Dockerfile#L4

masa213f (approved):
LGTM

PR #1536, opened 2025-02-14
Update pod-delete-rate-limiter for k8s 1.31

chez-shanpu:


pod-delete-rate-limiter/go.mod
```diff
@@ -1,70 +1,71 @@
 module github.com/cybozu/neco-containers/pod-delete-rate-limiter
 
-go 1.22.5
+go 1.23.0
```
YZ775, line 3: It might be that your local go version is old.
```suggestion
go 1.23.4
```

chez-shanpu: @YZ775 Thank you for your comment.
I've updated again using 1.23.6 but it's staying 1.23.0.
If there is no particular reason to specify the latest one, I think we can keep it as it is.

YZ775 (dismissed):
LGTM

chez-shanpu:
@YZ775 I forgot to update the Dockerfile. Could you check it again?

YZ775 (approved):
LGTM

PR #1538, opened 2025-02-14
Update etcd to 3.5.18

chez-shanpu:


etcd/Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Stage1: build from source
 FROM ghcr.io/cybozu/golang:1.22-jammy AS build
```
masa213f, line 4: Could you update base images?

masa213f (changes requested):

PR #1547, opened 2025-02-21
kube-state-metrics: Copy zoneinfo

masa213f:
This PR copies the zone info to the kube-stae-metrics's runtime image.

kube-stete-metrics v2.15.0 calls the [`time.LoadLocation()`](https://pkg.go.dev/time#LoadLocation) internally.
If the zone information does not exist in the runtime image, this function will fail and kube-state-metrics will panic.
To prevent this problem, I will copy the zone information.


YZ775 (approved):
LGTM

PR #1548, opened 2025-02-25
Fix: Make CI errors visible in GitHub UI

012xx:
## Overview
Fixed the issue so that errors can be checked directly from the GitHub UI.

## Features
Added Problem Matchers to highlight warnings (:warning::) and errors (:error::).
Removed incorrect behaviors that prevented proper error detection.

## Reason
The original implementation was meant to display error details in PR content during major version upgrades.
However, after updating to Go 1.24, we found that this was not working correctly. The CI was supposed to fail on a major version upgrade, but since the PR was never created, the errors couldn't be checked.
This fix ensures that even if the CI fails, the reason can still be seen in the GitHub UI.

## Ref
https://github.com/cybozu/neco-containers/actions/runs/13510755332/job/37750360290


rkwmt (approved):
LGTM

PR #1552, opened 2025-03-04
admission: fix log of PodCPURequestReducer

masa213f:
This PR fixes the log fields of PodCPURequestReducer.

It seems that `admission.Request.Name` is omitted in Pods created by DaemonSet, ReplicaSet, etc,.
In such cases, the fields in the log will be incomplete as follows.

```
{"level":"info","ts":"2025-03-04T03:45:46Z","logger":"reducer","msg":"patched","pod":"test/","container0":"from 100m to 50m"}
```

To deal with this, I fixes the webhook to output the prefix of the Pod name if the Name is empty.
The output example is as follows.

```
{"level":"info","ts":"2025-03-04T08:09:59Z","logger":"reducer","msg":"patched","namespace":"test","podName":"sts-0","container0":"from 100m to 50m"}
{"level":"info","ts":"2025-03-04T08:10:25Z","logger":"reducer","msg":"allowed","namespace":"test","podNamePrefix":"deploy-fb96c87c-","reason":"no cpu request"}
```

Note:

https://pkg.go.dev/k8s.io/api/admission/v1#AdmissionRequest

> // Name is the name of the object as presented in the request. On a CREATE operation, the client may omit name and
> // rely on the server to generate the name. If that is the case, this field will contain an empty string.
> // +optional
> Name string json:"name,omitempty" protobuf:"bytes,5,opt,name=name"

https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta

> // GenerateName is an optional prefix, used by the server, to generate a unique
> // name ONLY IF the Name field has not been provided.
> // If this field is used, the name returned to the client will be different
> // than the name passed. This value will also be combined with a unique suffix.
> // The provided value has the same validation rules as the Name field,
> // and may be truncated by the length of the suffix required to make the value
> // unique on the server.
> //
> // If this field is specified and the generated name exists, the server will return a 409.
> //
> // Applied only if Name is not specified.
> // More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#idempotency
> // +optional
> GenerateName string `json:"generateName,omitempty" protobuf:"bytes,2,opt,name=generateName"`


zeroalphat (approved):
LGTM

PR #1555, opened 2025-03-05
update: add Go 1.24 containers and update workflow configurations

012xx:
### Overview

update: add Go 1.24 containers and update workflow configurations

### What

-


golang-all/golang-1.24-jammy/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Go container
+
+FROM --platform=$TARGETPLATFORM ghcr.io/cybozu/ubuntu-dev:24.04
```
takahiro-yamada, line 3:
```suggestion
FROM --platform=$TARGETPLATFORM ghcr.io/cybozu/ubuntu-dev:22.04
```

golang-all/golang-1.24-jammy/README.md
```diff
@@ -0,0 +1,19 @@
+Go container
+============
+
+This directory provides a Dockerfile to build a Docker container
+that includes [Go](https://golang.org/) language runtime and following
+tools:
+
+* [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports)
+* [golint](https://github.com/golang/lint)
+* [ineffassign](https://github.com/gordonklaus/ineffassign)
```
takahiro-yamada, line 10: Add staticcheck here and in other READMEs please.

takahiro-yamada (changes requested):

golang-all/golang-1.24-jammy/README.md
012xx: Just to confirm if I understand correctly: 
The README previously did not mention staticcheck, but it was already being installed in the Dockerfile. 
Is it correct that the README was updated to eliminate this discrepancy?

takahiro-yamada: Yes, I asked you to add staticcheck because in Dockerfile it is installed but not mentioned in READMEs.
Thank you.

takahiro-yamada (approved):
LGTM

PR #1593, opened 2025-03-25
Set TMOUT on the teleport-node

rkwmt:


takahiro-yamada (approved):
LGTM

The use of `TMOUT` environment variable is valid according to `man bash` description.

```txt
[...]
       TMOUT  If set to a value greater than zero, TMOUT is treated as the default timeout for the read builtin.  The select command terminates if input does not arrive after TMOUT seconds when input is
              coming from a terminal.  In an interactive shell, the value is interpreted as the number of seconds to wait for input after issuing the primary prompt.  Bash terminates after waiting for
              that number of seconds if input does not arrive.
[...]
```

PR #1594, opened 2025-03-26
Update serf to 0.10.2.1

takara9:


serf/Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Stage1: build from source
 FROM ghcr.io/cybozu/golang:1.23-noble AS build
```
YZ775, line 4: Could you update base image too?
```suggestion
FROM ghcr.io/cybozu/golang:1.24-noble AS build
```

YZ775 (changes requested):

serf/Dockerfile
takara9: update

YZ775 (approved):
LGTM

PR #1596, opened 2025-03-27
Update bmc-log-collector 0.9.8

takara9:


zeroalphat (changes requested):
In go.mod, 1.24 is specified, but the Go image in the Dockerfile still seems to be 1.22.
Please update the Go image in the Dockerfile.

https://github.com/cybozu/neco-containers/blob/be8dfc9d67518e98d724dc4f5050dee2f4720a70/bmc-log-collector/Dockerfile#L2

zeroalphat (approved):
LGTM

PR #1602, opened 2025-04-04
lpp: update document

satoru-takeuchi:
Improve README and comments about device cleanup.


ushitora-anqou (changes requested):
The code changes look good to me, but could you write the commit message in English? (I'm not sure this is a strict rule or just a convention, though.)

ushitora-anqou (approved):

PR #1606, opened 2025-04-09
fix CVE-2025-31115

012xx:
### Overview 
- fix CVE-2025-31115
- [XZ Utils: Threaded decoder frees memory too early](https://github.com/tukaani-project/xz/security/advisories/GHSA-6cc8-p5mm-29w2)

### What
- Update TAG as detected by YAMORY


golang-all/golang-1.24-jammy/TAG
takahiro-yamada, line 1: According to Canonical, jammy is not affected by this CVE so you don't need to build this.
https://ubuntu.com/security/CVE-2025-31115

takahiro-yamada (changes requested):

golang-all/golang-1.24-jammy/TAG
012xx: I had misread the YAMORY notation. Thank you for pointing this out to me!

takahiro-yamada (approved):
LGTM

PR #1608, opened 2025-04-11
Rewrite diff checker

yokaze:
This PR rewrites diff checker.
stakater-reloader is updated to check the program works correctly.


chez-shanpu (approved):
LGTM

PR #1613, opened 2025-04-17
cilium: fix package dependency

yokaze:
This PR removes `libelf-dev:arm64` from the install script.
https://wiki.debian.org/Multiarch/HOWTO

This doesn't update `TAG` because the latest version was not created.


terassyi (approved):
LGTM

PR #1629, opened 2025-05-08
update rook to v1.17.1

cupnes:
Signed-off-by: Yuma Ogami <yuma-ogami@cybozu.co.jp>


rook/Dockerfile
peng225, line 1: The base image is not changed to Ceph v19.2.2.1 at the Stage2 build. Could you update it, too?

cupnes: I have addressed it.

peng225 (approved):

PR #1633, opened 2025-05-15
Reduce ephemeral storage request to 10Mi

masa213f:
This PR reduces the default resource requests of the local ephemeral storage added by admission from `200MiB` to `10MiB`.
In our environment, the local ephemeral storage is rarely used, and it is excessive even at 200 MiB.


zeroalphat (approved):
LGTM

PR #1640, opened 2025-05-30
Update meows runners to 0.20.2

gokken-roko:
meows released 0.20.2 : https://github.com/cybozu-go/meows/releases/tag/v0.20.2

## Changes

- Update meows-dctest-runner: runner image 0.20.2 
    - Update go version to 1.24.3
- Update meows-neco-runner: runner image 0.20.2
    - [meows already uses  ubuntu-24.04](https://github.com/cybozu-go/meows/blob/6444d8b02fa03385f0f56aa8aee7d9ecf64c8265/Dockerfile#L1). Thus, this runner updates to use 24.04.
    - Keep up with **[ubunu-debug-24.04](https://github.com/cybozu/ubuntu-base/blob/main/24.04/ubuntu-debug/Dockerfile)** updates.


meows-neco-runner/Dockerfile
```diff
@@ -15,10 +15,9 @@ LABEL org.opencontainers.image.source="https://github.com/cybozu/neco-containers
 USER root
 
 # Install debug tools.
-# https://github.com/cybozu/ubuntu-base/blob/main/22.04/ubuntu-debug/Dockerfile
+# https://github.com/cybozu/ubuntu-base/blob/main/24.04/ubuntu-debug/Dockerfile
```
takahiro-yamada, line 18: As you mentioned, the builder image of meows is 24.04 (`golang:1.24-noble`) but meows still depends on 22.04 as the controller and the runner images, so it seems better to keep 22.04 here too.

https://github.com/cybozu-go/meows/blob/6444d8b02fa03385f0f56aa8aee7d9ecf64c8265/Dockerfile#L7
https://github.com/cybozu-go/meows/blob/6444d8b02fa03385f0f56aa8aee7d9ecf64c8265/Dockerfile#L17

How do you think?

takahiro-yamada (changes requested):

meows-neco-runner/Dockerfile
gokken-roko: Thanks! You're right.

I should update all base images, not just the builder image.

In this PR, I will revert the 24.04 changes and try to update to 24.04 in the next release!

gokken-roko: Fixed!
https://github.com/cybozu/neco-containers/pull/1640/commits/197e0d8973a278d496f2406b9391177f2b7afe94

PR #1708, opened 2025-08-18
add websocket keepalive

terassyi:
- **Add websocket server/client**
- **Add metrics**
- **Add README**
- **Add maintenance procedures and image building parts**


websocket-keepalive/README.md
```diff
@@ -0,0 +1,62 @@
+# Websocket keepalive
+
+websocket-keepalive は websocket サーバ・クライアントの疎通の継続性を検証するためのプログラムである。
+
+## Server
```
tkna, line 5: [memo]
- The server accepts multiple clients at the same time.
- The server does not shutdown when a client closes its websocket connection.

websocket-keepalive/internal/server/server.go
```diff
@@ -0,0 +1,98 @@
...
+	conn.SetPingHandler(func(appData string) error {
+		slog.Debug("Received ping, sending pong", "message", appData, "remote_addr", conn.RemoteAddr().String())
+		return conn.WriteMessage(websocket.PongMessage, []byte("hello from server"))
+	})
+
+	// Message reading loop
+	for {
+		messageType, message, err := conn.ReadMessage()
+		if err != nil {
+			if websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway, websocket.CloseAbnormalClosure) {
```
tkna, line 83: `CloseNormalClosure` seems to be treated as "abnormal" closure.
The second argument of `websocket.IsUnexpectedCloseError` is "expectedCodes".

websocket-keepalive/internal/server/server.go
```diff
@@ -0,0 +1,98 @@
...
+
+	go func() {
+		slog.Info("WebSocket server starting", "addr", server.Addr)
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			slog.Error("Server failed to start", "error", err)
+		}
+	}()
+
+	<-stop
+	slog.Info("Shutting down server...")
```
tkna, line 52: It looks like websocket connections don't gracefully closed when shutdown from server side.
( ref: `conn.Close()` is not graceful)

websocket-keepalive/internal/server/server.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+			break
+		}
+
+		switch messageType {
+		case websocket.TextMessage:
+			slog.Debug("Received text message", "message", string(message), "remote_addr", r.RemoteAddr)
+		default:
+			slog.Debug("Received message", "type", common.WebSocketMessageType(messageType), "remote_addr", r.RemoteAddr)
+		}
```
tkna, line 96: This block is not used as
- `conn.ReadMessage` does not end when ping/pong
-  it breaks when a close message is received.

websocket-keepalive/internal/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+				retryCount = 0
+				waitingForReply = false
+			case <-done:
+				return
+			}
+		}
+	}()
+
+	go func() {
+		defer close(done)
```
tkna, line 145: It panics because of multiple close.

websocket-keepalive/internal/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+func RunWithConfig(config *Config, metricsConfig *metrics.Config) error {
+	u := url.URL{Scheme: "ws", Host: fmt.Sprintf("%s:%d", config.Host, config.Port), Path: "/ws"}
+
+	m, err := NewMetrics(metricsConfig)
+	if err != nil {
+		return err
+	}
+	if metricsConfig.Export {
+		slog.Info("Start metrics server", "listen", m.AddrPort)
+		go func() {
```
tkna, line 48: It looks like this goroutine never ends.

websocket-keepalive/internal/client/client.go
```diff
@@ -0,0 +1,173 @@
...
+						if err != nil {
+							slog.Error("Failed to send ping message", "error", err)
+							return
+						}
+						waitingCount = 0
+					} else {
+						// decide the connection is broken.
+						slog.Error("Reached to retry limit. The connection is broken.")
+						return
+					}
```
tkna, line 124: maybe it's better to reuse normal path code when retrying instead of using `else`.

websocket-keepalive/internal/client/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/VictoriaMetrics/metrics"
+	internalmetrics "github.com/cybozu/neco-containers/websocket-keepalive/internal/metrics"
+)
+
+type Metrics struct {
+	*internalmetrics.Metrics
+}
+
+var (
+	established    *metrics.Counter
```
tkna, line 13: Is there any particular reason to use `Counter` instead of `Gauge` for `established`?

websocket-keepalive/internal/common/utils.go
```diff
@@ -0,0 +1,18 @@
...
+		return "Binary"
+	case 8:
+		return "Close"
+	case 9:
+		return "Ping"
+	case 10:
+		return "Pong"
+	default:
+		return "Unknown"
+	}
```
tkna, line 17: To be strict, `0: Continuation Frame` is defined.

terassyi: I don't know there is type for 0. But in this package that I use doesn't implement 0.
https://github.com/gorilla/websocket/blob/v1.5.3/conn.go#L63

terassyi: done.

websocket-keepalive/internal/client/metrics.go
terassyi: No, I just thought `establsihed` is 1 or 0, so ether is fine.
Is Gauge better for this?

websocket-keepalive/internal/client/client.go
terassyi: remove this.

websocket-keepalive/internal/client/client.go
terassyi: It doesn't have to infinitely try to start metrics server, I found.
So, when it fails to start it , I changed to make this panic.

websocket-keepalive/internal/server/server.go
terassyi: Yes, this block is not used in normal path(ping/pong).
Do you think I should remove this?

PR #1710, opened 2025-08-20
Support testhttpd crossplatform

Till0196:


.github/workflows/main.yaml
```diff
@@ -234,6 +234,7 @@ jobs:
               container-image: "tempo"
             - dir: "./testhttpd"
               container-image: "testhttpd"
+              platforms: "linux/amd64,linux/arm64"
```
chez-shanpu, line 237: `load: "false"` has to be set when multi-platform build is used.

terassyi (approved):
LGTM

PR #1899, opened 2026-04-09
Build sealed secret form commit id

masa213f:
This PR updates sealed secrets 's docker file.
To retrieve the source code using the commit ID instead of the version tag.


YZ775 (approved):
LGTM

PR #1900, opened 2026-04-09
Build vault from commit id

masa213f:
This PR updates the vault's dockerfile to check out the vault source code by using the commit ID.


YZ775 (approved):
LGTM

PR #1910, opened 2026-04-13
Pin deps and Bump container image versions for bpftrace

chez-shanpu:


tkna (commented):
I think we should also update `BCC_TARGET`, `LIBBPF_TARGET`, and `BPFTRACE_TARGET` to use hashes instead of tags.

tkna (approved):
LGTM

PR #1917, opened 2026-04-14
update containers to support k8s 1.35

YZ775:


unbound/Dockerfile
```diff
@@ -5,8 +5,9 @@ ARG UNBOUND_VERSION=1.24.2
 WORKDIR /work
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN apt-get update && apt-get -y install --no-install-recommends libexpat1-dev libevent-dev
-RUN curl -sSLf https://nlnetlabs.nl/downloads/unbound/unbound-${UNBOUND_VERSION}.tar.gz | \
-    tar zxf - -C /work/
+RUN curl -sSLf -o unbound-${UNBOUND_VERSION}.tar.gz https://nlnetlabs.nl/downloads/unbound/unbound-${UNBOUND_VERSION}.tar.gz \
+    && echo "$(curl -sSLf https://nlnetlabs.nl/downloads/unbound/unbound-${UNBOUND_VERSION}.tar.gz.sha256)  unbound-${UNBOUND_VERSION}.tar.gz" | sha256sum -c - \
```
masa213f: Please use the hash value directly instead of downloading it.

When the binary is replaced for any reason, the SHA-256 file cannot be trusted either.

masa213f (changes requested):

masa213f (approved):
LGTM

PR #1919, opened 2026-04-14
squid: rebuild to use updated ubuntu image as base

hanapedia:
Signed-off-by: Hiroki Hanada <hiroki-hanada@cybozu.co.jp>


chez-shanpu (approved):
🚀

PR #1922, opened 2026-04-17
Securiy update for external-dns

terassyi:


hanapedia (approved):
LGTM

Maybe we could change the base image to scratch in the next regular update.
The upstream uses `ko` for building the image, which is essentially the same as using go  image for build stage and scratch for the final stage.
https://github.com/kubernetes-sigs/external-dns/blob/51995b83f4d64e128cf42ac4b9a94277d4cff7fc/Makefile#L108-L114

PR #1944, opened 2026-04-22
argo: add checksum verification

ystkfujii:
## Summary

- Pin all base images by digest with patch version tags
- Replace `git clone` and `curl` with `ADD --checksum` for source archives and binary downloads
- Use `--mount=type=bind` to avoid storing downloaded files in image layers
- Bump TAG patch: 1.9.0.2 / 3.3.6.2 / 1.1.1.2


argo-rollouts/Dockerfile
```diff
@@ -1,23 +1,26 @@
+# syntax=docker/dockerfile:1
```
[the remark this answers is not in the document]
ystkfujii: I wasn’t sure whether it was necessary to define it. I can address it if there is a request.

.github/actions/prepare_build_params/action.yaml
```diff
@@ -145,9 +145,11 @@ runs:
         # If base image is substituted by a variable, automatic detection is disabled
         # It instructs users to set scan-flag manually if missing
         if grep FROM Dockerfile | grep -F '$'; then
-          echo "!! FROM command may use a base image substituted by a variable."
-          echo "!! Please set scan requirement manually."
-          exit 1
+          if ! grep -qE 'FROM.*\$\{?TARGETARCH\}?' Dockerfile; then
```
ystkfujii: I added it because it causes an error without it.
I’m not sure why this condition is needed, but since:

- the container name is ultimately used, and
  - https://github.com/cybozu/neco-containers/blob/2d929d82e8913a252e340148f0ce2c7933b44708/.github/workflows/main.yaml#L26
- scanning is currently disabled
  - https://github.com/cybozu/neco-containers/pull/1876

I think this is fine for now.

012xx: ASK
Could you tell me how it fails if this change is not applied, just to be sure?

ystkfujii: The following line is detected, and the command exits with status 1, resulting in an error.

https://github.com/cybozu/neco-containers/blob/dc331fac25e3737554b291e1600df0b5c27b3756/argocd-image-updater/Dockerfile#L20

plz see below for the error details
- https://github.com/cybozu/neco-containers/actions/runs/24776171655/job/72494889000

ystkfujii: For clarification,

> I’m not sure why this condition is needed

Here, “the condition” refers to the original logic, 

```
if grep FROM Dockerfile | grep -F '$'; then
```

not the logic I added.

```
if ! grep -qE 'FROM.*\$\{?TARGETARCH\}?' Dockerfile; then
```

argocd/Dockerfile
```diff
@@ -1,42 +1,58 @@
+# syntax=docker/dockerfile:1
 # Argo CD container
 
+ARG ARGOCD_VERSION=3.3.6
+ARG KUSTOMIZE_VERSION=5.8.1
```
pddg: kustomize is downgraded here
https://github.com/cybozu/neco-containers/pull/1945

.github/actions/prepare_build_params/action.yaml
012xx: Is the reason for making $TARGETARCH the only exception because it is a standard Docker Buildx variable that is automatically replaced based on the architecture, such as linux/amd64 or linux/arm64, during the build process, and thus it is considered not to be an issue?

ystkfujii: About the original logic for setting the scan flag, it looks like the flag is not actually used right now. It was originally used for Trivy scanning, but since Trivy scan is currently disabled, I do not think this is an issue at the moment. It might become a problem if Trivy scan is re-enabled in the future, though.

There are still a lot of unknowns here, so I think it is better to revisit that when Trivy scan is actually brought back. Even if something does break, it would most likely only affect the Trivy scan result or the Yamory integration, so I do not think it is critical.

ystkfujii: Ah... While writing this comment, I started to think that simply setting `request-scan: false` might be a better approach. I will update it that way.

thanks !

argocd/Dockerfile
ystkfujii: Good catch, thank you for the comment!

.github/actions/prepare_build_params/action.yaml
ystkfujii: i fixed it

https://github.com/cybozu/neco-containers/pull/1944/changes/07497627ba8ba2698763d9ac96bfac4f2b0dff4e

012xx: Thank you. That is very clear!
I understand that Trivy is currently stopped, but since the scan should resume soon, we will reconsider it then.

012xx (approved):
LGTM🚀

PR #1948, opened 2026-04-28
rook: update to v1.19.4

molpako:


ushitora-anqou (changes requested):
The file changes looks good to me, but the message of the first commit looks wrong. It should include "Revert" at the beginning?

molpako:
Sorry, I meant to write `remove patch ~` there!

ushitora-anqou (approved):

PR #1959, opened 2026-05-07
feat: Add subnamespace deletion validator

ystkfujii:


ystkfujii:
dup: https://github.com/cybozu/neco-containers/pull/2027

PR #1966, opened 2026-05-12
lock go installed tools versions

hanapedia:
- **cilium: lock delve version**
- **setup: lock container-tag-exists version**


chez-shanpu (approved):
LGTM

PR #1975, opened 2026-05-14
kubernetes: split into per-minor-version directories

zeroalphat:
This PR reorganizes the Kubernetes image build definitions by Kubernetes minor version.
This PR does not update any TAG versions and focuses solely on reorganizing the Kubernetes images. Therefore, no Kubernetes images will be updated.


kubernetes/k8s-1.33/README.md
```diff
@@ -0,0 +1,31 @@
...
+- kube-scheduler
+- kubelet
+
+Usage
+-----
+
+To launch `apiserver` by `docker run`:
+
+    $ docker run --net=host --name apiserver -d \
+        ghcr.io/cybozu/kubernetes:1.33 kube-apiserver \
```
YZ775, line 20: The README.md file is rarely changed except for the version tag. So how about placing this file under `kubernetes/README.md` instead of `kubernetes/k8s-x.y.z/README.md`?

YZ775 (changes requested):
Cloud you update maintenance.md too?
https://github.com/cybozu/neco-containers/blob/main/maintenance.md#kubernetes

kubernetes/k8s-1.33/README.md
zeroalphat: I agree that updates to README.md are infrequent, but I wasn't quite sure about the reasoning for placing it in the kubernetes directory. Since directories that maintain multiple versions, like golang-all, each have their own README.md, I think it would be better for kubernetes to follow that same style.

YZ775: Sorry, I didn't know that golang-all has same structure.
Since it's important to maintain consistency in the structure, I think it was fine to leave it as is.

YZ775 (approved):
LGTM

PR #1979, opened 2026-05-15
cilium: rebuild containers

yokaze:
We need to rebuild the containers because the previous release attempt failed due to CI instability.


hanapedia (approved):
LGTM

PR #1987, opened 2026-05-28
Declare AGPLv3 as the chosen license from Redis v8 tri-license

takahiro-yamada:
Declare the AGPLv3 license and include NOTICE for redis image.


012xx (approved):
LGTM

PR #1989, opened 2026-05-29
Specify kindest node digest

hanapedia:
- specify image digest for kindest/node
- use aqua for installing tools for e2e

test build with TAG bump: https://github.com/cybozu/neco-containers/pull/1990


terassyi (approved):
Thanks!

PR #2001, opened 2026-06-10
Update pomerium to 0.32.8

pddg:
- Update pomerium to latest version
- Follow upstream's node version


pomerium/TAG
```diff
@@ -1 +1 @@
-0.32.7.0
+0.32.8.0
```
gokken-roko:
```suggestion
0.32.8.1
```

I think container image version is start from `1`.

https://github.com/cybozu/neco-containers#tag-naming

> For example, if the upstream version is `X.Y.Z`, the first image for this version will
be tagged as `X.Y.Z.1`.

gokken-roko (changes requested):

pomerium/TAG
pddg: I'm wondering why it starts from 1, as I think it should start from 0, but I'll fix it here for now.
https://github.com/cybozu/neco-containers/pull/2001/commits/242b8a45d8c90fa558b9f944de42c5f48ee08404

gokken-roko: I don't know the exact reason. (`.1` emphasizes "the first container build of the image", maybe.)
Anyway, I think starting from `1` is the better choice to keep consistency for now 🙏

gokken-roko (approved):
LGTM

PR #2005, opened 2026-06-12
Fix Update Go Versions GitHub Actions workflow for sha256 checksum

gokken-roko:
## Changes

- Refactor golang-all Dockerfiles to set go version and sha256 checksum easily.
- Update `.github/workflows/update.yaml` to update sha256 checksums automatically.
    - Currently, this workflow only updates the go version, and the workflow is failing.
    - ref. https://github.com/cybozu/neco-containers/actions/runs/27381984110  

## Test

I triggered the workflow manually and confirmed success.
https://github.com/cybozu/neco-containers/actions/runs/27396138312/job/80963708515


rkwmt (approved):
LGTM

PR #2014, opened 2026-06-18
Refactoring build matrix

pddg:
This PR refactors the build matrix generation logic to facilitate future feature additions.
I identified the following issues with the previous implementation:

- The build matrix was defined as inline strings within the workflow file.
- It was difficult to verify the exact JSON structure generated by the matrix.
- Complex custom actions had inputs/outputs containing confusing values (e.g., push vs. docker_push).
- Scan targets were selected based on heuristic judgments.

To address these, I have implemented the following changes:

- Separated the build matrix into a standalone file ( `.github/build-targets.yaml`).
- Created a script (`build_decision`) that takes the file path and an image name as input to output the necessary build parameters.

The build_decision script merges parameters defined in the build matrix with several automatically generated ones. This makes it easier to pass parameters into the build pipeline and extend them.
This refactor is intended to simplify extending scan-related behaviors for each image in the future.

```
❯ SKIP_TAG_CHECK=true ./build_decision .github/build-targets.yaml cilium | jq .
SKIP_TAG_CHECK is true; skipping tag existence check and building anyway
{
  "dir": "./cilium",
  "container-image": "cilium",
  "runner": "neco-containers-ubuntu-22.04",
  "go-version": "1.23.3",
  "make-targets": "checkout status test-e2e-setup test-e2e-upgrade-inotify",
  "make-post-targets": "test test-e2e",
  "load": true,
  "scan": {
    "enabled": true,
    "image": "ghcr.io/cybozu/cilium",
    "tag": "1.17.12.3"
  },
  "build": true,
  "tags": [
    "ghcr.io/cybozu/cilium:1.17.12.3",
    "ghcr.io/cybozu/cilium:1.17"
  ],
  "platforms": "linux/amd64"
}
```

### Breaking Changes

- `request-scan` is renamed to `scan.enabled`, and the `scan.enabled` flag is now mandatory in the build matrix.
- Even if files related to the target container image are modified, job generation itself will now be skipped if an image has already been pushed with that `TAG`.
- Unneeded actions are removed
  - prepare_build_params
  - prepare_matrix

In order to suppress the generation of jobs themselves, the build matrix generation job now checks for the existence of images. In the worst-case scenario, it would verify whether all images exist in serial. However, typically such situations are rare, and I believe the benefits of being able to skip the job generation itself are greater.

---

I have also included other minor fixes:

- Reduced inputs by using `${{ github.token }}` instead of passing secrets.GITHUB_TOKEN.
- Skip QEMU setup when it is unnecessary.


build_decision
```diff
@@ -0,0 +1,113 @@
...
+if [ -z "${ENTRY}" ]; then
+  echo "Error: IMAGE '${IMAGE_NAME}' not found in MATRIX_FILE '${MATRIX_FILE}'" >&2
+  exit 1
+fi
+
+# === Validate required fields ===
+
+DIR=$(echo "${ENTRY}" | jq -r '.dir')
+if [ ! -d "${DIR}" ]; then
+  echo "Error: DIR '${DIR}' does not exist or is not a directory" >&2
```
tkna, line 59: `ENTRY` can be multiple objects when multiple entries exist in `build-targets.yaml` for a container image name(e.g. golang), so in that case this script fails here.

.github/actions/build_envoy/action.yaml
```diff
@@ -107,17 +116,17 @@ runs:
...
+        tags: ${{ join(fromJSON(steps.prepare.outputs.plan).tags, ',') }}
         load: true
     - name: Test image
-      if: ${{ steps.prepare.outputs.build }}
+      if: steps.prepare.outputs.build == 'true'
       shell: bash
-      run: docker run --rm ${{ steps.prepare.outputs.tag }} envoy --version
+      env:
+        IMAGE: ${{ fromJSON(steps.prepare.outputs.plan).tags[0] }}
+      run: docker run --rm "$IMAGE:$TAG" envoy --version
```
tkna: `.tags[0]` contains a tag so `$IMAGE:$TAG` would be an invalid tag reference.

build_decision
pddg: Ah... You are right. I will fix the arguments and validation logic.

pddg: I have made it so that they are uniquely identified using a combination of `dir` and `container-image`. I decided not to include `target` (currently, there are no cases where the `target` is different but the `container-image` is the same, and such a scenario seems unlikely).
https://github.com/cybozu/neco-containers/pull/2014/commits/d487b957e50d2475af7bd3e54419075932e61278

.github/actions/build_envoy/action.yaml
pddg: Fixed and add `--pull=never` to use local image.
https://github.com/cybozu/neco-containers/pull/2014/commits/a1bd61132a19131f4dccf97f7464e6f5dba65a63

pddg:
After the logic review is finished, I will rebase this branch and incorporate the changes to the build matrix in the main branch.

build_decision
pddg:
```
❯ SKIP_TAG_CHECK=true ./build_decision ".github/build-targets.yaml" golang ./golang-all/golang-1.25-jammy 2>/dev/null | jq .
{
  "dir": "./golang-all/golang-1.25-jammy",
  "container-image": "golang",
  "platforms": "linux/amd64,linux/arm64",
  "scan": {
    "enabled": true,
    "image": "ghcr.io/cybozu/golang",
    "tag": "1.25.11.2_jammy"
  },
  "build": true,
  "tags": [
    "ghcr.io/cybozu/golang:1.25.11.2_jammy",
    "ghcr.io/cybozu/golang:1.25-jammy"
  ],
  "load": false
}
```

tkna (commented):
LGTM

tkna (dismissed):

tkna (dismissed):

pddg:
I fixed an issue where some untracked files were accidentally included in a commit during a rebase.

tkna (approved):

PR #2018, opened 2026-06-22
Enable container scan

pddg:
## Summary

Scanning of container images has been frozen since the Trivy incident. However, due to an update to our internal policy, it has become available again, so I will implement it in a way that complies with the new policy.

This PR adds the following actions:

- `setup-cosign`: Installs the cosign command
- `setup-trivy`: Retrieves the Trivy binary and verifies the signature using cosign
- `setup-yamory-scripts`: Retrieves the scripts for uploading to yamory and performs checksum verification

I will integrate the scanning workflow using these components.

In the previous implementation, scanning was implemented within the `build_container` action. This meant that if the container build and push succeeded but the scan failed, retrying would skip the build, sometimes making it difficult to retry only the scan.
In this implementation, I will separate the steps for building and pushing the container image from the step for scanning the pushed container image. This provides two benefits:

- Retries can be performed for the scan alone without waiting for the time-consuming container image build
- Minimal permissions can be granted to the container image scanning job, which frequently uses external binaries

`build_decision` now outputs as follows:

```
❯ SKIP_TAG_CHECK=true ./build_decision .github/build-targets.yaml golang ./golang-all/golang-1.25-jammy 2>/dev/null | jq .
{
  "dir": "./golang-all/golang-1.25-jammy",
  "container-image": "golang",
  "platforms": "linux/amd64,linux/arm64",
  "scan": {
    "enabled": true,
    "team": "NECO",
    "identifier": "ghcr.io/cybozu/golang:1.25-jammy",
    "is_public": false,
    "image": "ghcr.io/cybozu/golang",
    "tag": "1.25.11.2_jammy"
  },
  "build": true,
  "tags": [
    "ghcr.io/cybozu/golang:1.25.11.2_jammy",
    "ghcr.io/cybozu/golang:1.25-jammy"
  ],
  "load": false
}
```
 
## `scan.team` is required to enable scanning

In addition, multiple teams are collaborating in this repository. The scan results for container images should also be pushed to separate workspaces for each team.

I have made the `scan.team` specification mandatory when `scan.enabled=true` in `build-targets.yaml`. If you wish to enable container image scanning, you must specify which team the scan results should be pushed to.

## Specifying the Image Identifier Manually

In Yamory, image scan results are identified by a unique identifier, and different identifiers are registered as distinct assets. Previously in this repository, for images other than golang, the container image URL excluding the tag name was used as the identifier ([ref](https://github.com/cybozu/neco-containers/blob/f1d801bbaf700a57a47b49259370b790f05efcee/.github/actions/trivy_scan/action.yaml#L33-L39)), and we will basically follow this approach. As an exception, for images that maintain multiple versions over a long period (golang, kubernetes), we have made it possible to specify the identifier manually.

The reason we do not automatically retrieve the value from the `BRANCH` file is that the `BRANCH` file can contain multiple values (prev is [ref](https://github.com/cybozu/neco-containers/blob/f1d801bbaf700a57a47b49259370b790f05efcee/.github/actions/prepare_build_params/action.yaml#L81-L89), current is [ref](https://github.com/cybozu/neco-containers/blob/344ecd36fa27d71fafc6c8c6808439dfbef1b62c/tags_gen#L36-L41)), making it impossible to determine which one should be used as the identifier.

## Specifying Secrets Dynamically

The `scan.team` values specified in `build-targets.yaml` each have different secrets. Therefore, it is necessary to retrieve these dynamically when executing a scan. There are several ways to achieve this:

- Using Environment Secrets
- Dynamically assembling the `*` part of `secrets.TOKEN_*`

In this instance, I chose the latter. I decided not to use "Environment" because it felt slightly outside its intended purpose. That said, I believe this is a point of contention, and I would welcome your feedback.

[It has been verified](https://github.com/pddg/actions-playground/actions/runs/27997400554/job/82862374839) that secrets can be specified using a format like `secrets[format("SECRET_{0}"), matrix.value)]`. If it does not exist, it simply results in an empty string and does not fail. Therefore, the value is validated and rejected within the receiving workflow.

## Token should be added to secrets

For this time, I have configured it to push all images to the `neco` team's workspace as usual. Therefore, it is necessary to add `YAMORY_TOKEN_TEAM_NECO` before merging.

Additionally, when pushing to other teams' workspaces, we will need to add a token along with changing the team setting.


.github/actions/setup-cosign/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: Setup cosign
```
tkna, line 1: There is a typo in the file name.

.github/actions/setup-trivy/action.yaml
```diff
@@ -0,0 +1,61 @@
+name: Setup Trivy
```
tkna, line 1: There is a leading space in the filename.

.github/actions/build_envoy/action.yaml
```diff
@@ -18,7 +18,16 @@ inputs:
...
+  image:
+    description: "Built image name"
+    value: ${{ steps.prepare.outputs.image }}
+  tag:
+    description: "Built image tag"
+    value: ${{ steps.prepare.outputs.tag }}
+  image-identifier:
+    description: "Built image identifier"
+    value: ${{ steps.prepare.outputs.image-identifier }}
+  changed :
```
tkna:
```suggestion
  changed:
```

.github/actions/setup-trivy/ action.yaml
```diff
@@ -0,0 +1,61 @@
...
+        echo "::endgroup::"
+
+        echo "::group::Check the commit is signed"
+        SOURCE_HASH=`cat bundle.json | \
+          jq -r '.dsseEnvelope.payload' | base64 -d | \
+          jq -r '.predicate.buildDefinition.resolvedDependencies[] | select(.uri | startswith("git+https://github.com/aquasecurity/trivy@refs/tags/")) | .digest.gitCommit'`
+        echo ${SOURCE_HASH}
+        curl -s -H "Accept: application/vnd.github+json" \
+          "https://api.github.com/repos/aquasecurity/trivy/commits/${SOURCE_HASH}" | \
+          jq '.commit.verification'
```
tkna: Is it OK to just display the verification result rather than exit 1 even if the verification is failed?

.github/actions/setup-trivy/action.yaml
```diff
@@ -0,0 +1,61 @@
...
+          --bundle "${ARTIFACT}.sigstore.json" \
+          --certificate-oidc-issuer "https://token.actions.githubusercontent.com" \
+          --certificate-identity "https://github.com/aquasecurity/trivy/.github/workflows/reusable-release.yaml@refs/tags/v${TRIVY_VERSION}" \
+          "${ARTIFACT}"
+        echo "::endgroup::"
+
+        # https://docs.yamory.io/35d9ca090120803f8b51d6a50d6ce3e0#block-36b9ca090120808ca7fef8760e2b106c
+        echo "::group::Check Provenance"
+        DIGEST=$(sha256sum "${ARTIFACT}" | cut -d' ' -f1)
+        curl -L "https://api.github.com/repos/aquasecurity/trivy/attestations/sha256:$DIGEST" | \
```
tkna, line 34: Github API is not authorized. (may reach the rate limit)

.github/build-targets.yaml
```diff
@@ -324,18 +377,22 @@
   container-image: "pomerium"
   scan:
     enabled: true
+    team: neco
 - dir: "./prometheus-adapter"
   container-image: "prometheus-adapter"
   scan:
     enabled: false
+    team: neco
```
tkna: `team: neco` is added even if `enabled: false` only for `prometheus-adapter`. Is there a specific purpose?

.github/actions/setup-trivy/ action.yaml
```diff
@@ -0,0 +1,61 @@
...
+        echo "::endgroup::"
+
+        echo "::group::Install trivy"
+        tar -xzf "${ARTIFACT}" -C /tmp
+        chmod +x /tmp/trivy
+        mv /tmp/trivy /usr/local/bin/
+        rm -f "${ARTIFACT}" "${ARTIFACT}.sigstore.json" bundle.json /tmp/trivy
```
tkna: `/tmp/trivy` (L58) is not supposed to exist because it has already moved.

.github/actions/setup-trivy/action.yaml
pddg: This is provided as the official procedure, and conversely, it is known that [problems](https://github.com/aquasecurity/trivy/discussions/10441) occur when authorization is implemented. Therefore, unless rate-limiting issues become apparent, the policy is to operate it as guided for the time being.

tkna:
> In this instance, I chose the latter. I decided not to use "Environment" because it felt slightly outside its intended purpose. That said, I believe this is a point of contention, and I would welcome your feedback.

I think it's reasonable to use the latter approach. As far as I investigated, `Environment` secret is more secure since it can set branch protection, but its not complete in terms of privilege separation because repository admin can create/edit all environments.

.github/actions/build_envoy/action.yaml
pddg: fixed https://github.com/cybozu/neco-containers/pull/2018/commits/79302d798b00a6aeebc902a80de169c876247516

.github/actions/setup-trivy/action.yaml
pddg: fixed https://github.com/cybozu/neco-containers/pull/2018/commits/79302d798b00a6aeebc902a80de169c876247516

.github/actions/setup-cosign/action.yaml
pddg: fixed https://github.com/cybozu/neco-containers/pull/2018/commits/79302d798b00a6aeebc902a80de169c876247516

.github/build-targets.yaml
pddg: fixed https://github.com/cybozu/neco-containers/pull/2018/commits/11fcfa4bf44136a17812148d79185617b2edaba8

.github/actions/setup-trivy/ action.yaml
pddg: fixed https://github.com/cybozu/neco-containers/pull/2018/commits/58d4a1dfbef1ad020f3ca5ca3b42177110264f04

.github/actions/setup-trivy/ action.yaml
pddg: This was originally a direct copy-paste of the official procedure, which seemed to be intended for manual visual inspection by a human. I have updated it to check only whether a valid signature is present during the CI process. https://github.com/cybozu/neco-containers/pull/2018/commits/889bb9d257a5425df8baf2a4add46777fa0df71d

tkna (approved):
LGTM

pddg:
@tkna 
Can I leave the registration of the YAMORY_TOKEN_TEAM_NECO secret to you? (I do not have the necessary permissions). Once that is complete, I would like to proceed with the merge.

tkna:
@pddg I have registered `YAMORY_TOKEN_TEAM_NECO` as a repository secret.

PR #2023, opened 2026-06-26
Remove .gitignore and commit crds for admission

takahiro-yamada:
This PR removes the `.gitignore` under `admission/config/crd/bases`, so that generated CRD manifests are committed to the repository.

The reason why the generated CRDs were ignored is unknown. After investigation, I could not find any documented rationale.

However, I believe there's no longer a need to keep ignoring them:

The CRD files are not large. Committing them lets us review CRD changes in PR diffs.


rkwmt (approved):
LGTM

PR #2027, opened 2026-06-30
[neco-admission] Add namespaceDeletionValidator

ystkfujii:
## Overview

This PR adds a ValidatingWebhook that denies Namespace deletion when protected resources still exist in the Namespace.

Currently, resources annotated with `admission.cybozu.com/prevent: delete` are protected from direct DELETE requests. However, when the Namespace itself is deleted, Kubernetes starts deleting resources in the Namespace. In that process, deletion of protected resources may be denied, but RBAC resources such as Roles and RoleBindings can still be deleted.

As a result, the Namespace can get stuck in the Terminating state. The protected resources remain, but the RBAC resources in the Namespace are gone, so non-admin users can no longer recover the Namespace by themselves.

This PR checks for protected resources when a Namespace DELETE request is made. If any configured protected resource has the `admission.cybozu.com/prevent: delete` annotation, the webhook denies the Namespace deletion before the deletion process starts. This prevents RBAC resources from being deleted first and leaving the Namespace in an unrecoverable state for non-admin users.

## Example

If a protected resource has the `admission.cybozu.com/prevent: delete` annotation, Namespace deletion is denied.

```console
$ kubectl delete ns sb
Error from server (Forbidden): admission webhook "vnamespacedeletion.kb.io" denied the request: Service sb/sbb is protected from deletion
```

## RBAC

This PR does not include RBAC changes for the webhook itself.

The new webhook lists resources configured in `protectedResources` inside the target Namespace. Depending on which resources are configured, additional RBAC permissions for the webhook may be required.


012xx (approved):
LGTM

PR #2030, opened 2026-07-02
chore: Bump up cAdvisor

ystkfujii:
https://github.com/google/cadvisor/compare/v0.56.2...v0.57.0

https://github.com/google/cadvisor/compare/v0.57.0...v0.60.3


cadvisor/Dockerfile
```diff
@@ -1,16 +1,23 @@
...
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 WORKDIR /go/src/github.com/google/cadvisor
-RUN curl -fsSL -o cadvisor.tar.gz "https://github.com/google/cadvisor/archive/${CADVISOR_COMMIT}.tar.gz" \
-    && tar -x -z --strip-components 1 -f cadvisor.tar.gz \
-    && rm -f cadvisor.tar.gz \
+COPY --from=source /cadvisor.tar.gz cadvisor.tar.gz
+RUN tar -x -z --strip-components 1 -f cadvisor.tar.gz \
     && cd cmd \
     && CGO_ENABLED=0 go build -tags netgo -ldflags="-w -s" -o cadvisor .
```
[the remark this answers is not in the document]
ystkfujii: Deleting it may actually increase the size.

> Each layer is only a set of differences from the layer before it. Note that both adding, and removing files will result in a new layer. 

https://docs.docker.com/engine/storage/drivers/

cadvisor/Dockerfile
```diff
@@ -1,16 +1,23 @@
...
-# Stage1: build from source
-FROM ghcr.io/cybozu/golang:1.25-noble AS build
+# Stage0: download source archive
+FROM scratch AS source
+
+ARG CADVISOR_VERSION=v0.57.0
+ARG CADVISOR_TARBALL_SHA256=2f700ada18220d6da193e6ecf2d35a52b84e4c38bc20aa08b2e5e3a17f68d1f5
 
-ARG CADVISOR_COMMIT=7cbdee9260c4292d5d27d00416ce8a7dc1dc12a0
+ADD --checksum=sha256:${CADVISOR_TARBALL_SHA256} \
```
YZ775: Since only the binary files are copied in the final stage, I think there was no need to split the download into separate stages.

cadvisor/TAG
```diff
@@ -1 +1 @@
-0.56.2.2
+0.57.0.1
```
YZ775: Latest version of cadvisor is 0.60.3.
https://github.com/google/cadvisor/releases

YZ775 (changes requested):

cadvisor/Dockerfile
ystkfujii: I agree that the separate source stage may not be necessary, since the final stage only copies the binary and license.

My intention was to separate source acquisition/checksum verification from the build stage, but I don’t have a strong opinion here.

Would you prefer merging the download step into the build stage for simplicity?

cadvisor/TAG
ystkfujii: Done

https://github.com/cybozu/neco-containers/pull/2030/changes/0b6f05e5e007fa9613e25a02ae3e693bb9fe768f

cadvisor/Dockerfile
ystkfujii: https://github.com/cybozu/neco-containers/pull/2030/changes/51185d5986eb3f4c90e3f05632a87325f8031ed8

YZ775 (approved):
Thank you!
LGTM

PR #1915, opened 2026-04-13
testhttpd: Update Makefile with commit hash SHAs

morimoto-cybozu:
See-also: https://github.com/cybozu-private/neco-tasks/issues/807
See-also: https://github.com/cybozu-private/neco-tasks/issues/987
Signed-off-by: morimoto-cybozu <kenji_morimoto@cybozu.co.jp>

PR #2031, opened 2026-07-02
chore: Bump up testhttpd

ystkfujii:
previous pr : https://github.com/cybozu/neco-containers/pull/1915


testhttpd/src/go.mod
```diff
@@ -1,13 +1,15 @@
 module github.com/cybozu/neco-containers/testhttpd/src
 
-go 1.26.1
```
ystkfujii, line 3: We can only update Go up to 1.26.3.

- https://github.com/cybozu/neco-containers/blob/39de8fc8f16622e2edcd6fa7fe2dbb3dc428cf27/.github/workflows/main.yaml#L48
- https://github.com/cybozu/neco-containers/actions/runs/28570510947/job/84707028059#step:5:31

It would be possible to support this by defining `matrix.job.go-version`, but I’m thinking we don’t need to handle it for now. Please let me know if you think we should.

testhttpd/src/Makefile
```diff
@@ -4,11 +4,6 @@ SUDO_GO = $(SUDO) $(shell which go)
 .PHONY: all
 all: check-generate test
 
-.PHONY: setup
-setup:
```
ystkfujii, line 8: Migrated to `go tool`, so we no longer need to manage hashes !

testhttpd/src/Makefile
```diff
@@ -17,7 +12,12 @@ check-generate:
...
-	test -z "$$(custom-checker -restrictpkg.packages=html/template,log ./... 2>&1 | tee /dev/stderr)"
+	go tool staticcheck ./...
+	test -z "$$(go tool custom-checker -restrictpkg.packages=html/template,log ./... 2>&1 | tee /dev/stderr)"
 	go vet ./...
 	$(SUDO_GO) test -race -v ./...
+
+.PHONY: update-tool
+update-tool:
+	go get tool
+	go mod tidy
```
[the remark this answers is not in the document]
ystkfujii: > the tool meta-pattern resolves to the list of tools

https://go.dev/ref/mod#go-mod-file-tool

morimoto-cybozu (approved):

PR #2033, opened 2026-07-03
reguler-update: update dependencies

YZ775:


s3gw/go.sum
masa213f, line 1: Please update dependencies in Makefile as well.
https://github.com/cybozu/neco-containers/blob/6edb20ca54fb7dd60347beea2367670aa378c031/s3gw/Makefile#L9

masa213f (changes requested):

s3gw/go.sum
YZ775: Thank you.
Changed the Makefile to use go tool.

masa213f (approved):
LGTM

PR #2036, opened 2026-07-06
refactor: use resource name for namespace deletion validator config

ystkfujii:
Changed protectedResources in NamespaceDeletionValidator from kind-based configuration to resource-based configuration. This aligns the configuration with ValidatingAdmissionPolicy.

- https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/

The validator now resolves the configured resources and checks list permissions at startup.


rkwmt (approved):
LGTM

PR #2041, opened 2026-07-09
external-dns: regular-update & use scratch base image

hanapedia:
- build external-dns v0.21.0
- use scratch as base image


terassyi (approved):
LGTM

PR #2048, opened 2026-07-09
machines-endpoints: Fix to handle >1000 machines

morimoto-cybozu:
This commit fixes machines-endpoints to handle more than 1000 machines
by splitting endpoints into multiple EndpointSlice objects.
EndpointSlice API does not accept more than 1000 endpoints in a single
object.

This commit also drops support for Endpoints API.


machines-endpoints/pkg/machines-endpoints/main.go
```diff
@@ -206,62 +212,91 @@ func (c client) updateTargetEndpoints(ctx context.Context, targetIPs []net.IP, t
...
-	ep := &corev1.Endpoints{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: target,
-			Labels: map[string]string{
-				"endpointslice.kubernetes.io/skip-mirror": "true",
+	// Create or update EndpointSlice(s)
+	endpointSliceInterface := c.k8s.DiscoveryV1().EndpointSlices(ns)
+	sliceIndex := 0
+	sliceNames := []string{}
+	for chunkedIPs := range slices.Chunk(targetIPs, maxEndpoints) {
```
YZ775, line 219: The order of `targetIPs` is unstable because it is added from unsorted sabakan query result.
I think it's better to sort `targetIPs` because contents of the EndpointSlices could be changed between the reconcile loop of the machines-endpoints.

https://github.com/cybozu/neco-containers/blob/90d6a160d2ef035a42fa0b6e301f85e93a474b51/machines-endpoints/pkg/machines-endpoints/main.go#L459

YZ775 (approved):
LGTM

PR #2049, opened 2026-07-10
external-dns: pin builder image and verify downloaded tar

hanapedia:
Added security checks missed in the previous PR. https://github.com/cybozu/neco-containers/pull/2041


yokaze (approved):
LGTM

PR #2051, opened 2026-07-13
Add CODEOWNERS and CI check for ownership coverage

shunki-fujita:
## Summary
- Add `.github/CODEOWNERS` that assigns each top-level product directory to its owning team, with `* @cybozu/neco` as the default fallback owner.
- Add a GitHub Actions workflow (`.github/workflows/check-codeowners.yaml`) that fails when a top-level directory has no `/<dir>/` entry in `.github/CODEOWNERS`, so new directories cannot be merged without an explicit owner.
- Document usage notes in the CODEOWNERS header comment.
- Split `.github/build-targets.yaml` into per-directory files under `.github/build-targets/` (one YAML file per top-level directory). `generate_matrix` merges them at runtime, preserving the existing build flow. The field schema documentation is moved to a header comment in `generate_matrix`.

## Test plan
- [x] Confirm the `check-codeowners` workflow succeeds on this PR.
- [x] Confirm the workflow fails when a top-level directory without an entry is added (verified locally, see [comment](https://github.com/cybozu/neco-containers/pull/2051#issuecomment-4955295527)).
- [x] Sanity-check ownership assignments for each team.
- [x] Confirm the split build-targets files produce the same merged matrix as the original (114 entries, `(dir, container-image)` set identical).
- [x] Confirm the `main` workflow still filters and builds the expected images on this PR.


shunki-fujita:
I confirmed that the CI script fails when an unowned directory is added locally.

```console
$ mkdir __test-noowner
$ bash -c '
set -euo pipefail
missing=()
for d in */; do
  name="${d%/}"
  if ! grep -qE "^/${name}/" .github/CODEOWNERS; then
    missing+=("$name")
  fi
done
if [ ${#missing[@]} -gt 0 ]; then
  {
    echo "::error::The following top-level directories are missing from .github/CODEOWNERS:"
    printf " - %s\n" "${missing[@]}"
  } >&2
  exit 1
fi
'
::error::The following top-level directories are missing from .github/CODEOWNERS:
 - __test-noowner
$ echo $?
1
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,123 @@
...
+/opentelemetry-collector/         @cybozu-private/landsat
+/prometheus-config-reloader/      @cybozu-private/landsat
+/pushgateway/                     @cybozu-private/landsat
+/pyroscope/                       @cybozu-private/landsat
+/tempo/                           @cybozu-private/landsat
+/victoriametrics/                 @cybozu-private/landsat
+/victoriametrics-operator/        @cybozu-private/landsat
+
+# ---- CSA (@cybozu-private/csa) ----
+/bucket-provisioner-light/        @cybozu-private/csa
```
toshipp: This was not added by CSA.

ystkfujii: PDX

takahiro-yamada:
The CODEOWNERS assignments for PDX look fine except for this https://github.com/cybozu/neco-containers/pull/2051#discussion_r3569309395.

takahiro-yamada (approved):
LGTM. No concerns regarding the PDX-related changes.

PR #2052, opened 2026-07-15
neco-exporter: implement kubelet config exporter

hanapedia:
- Add a new kubelet node-scoped collector that reads kubelet's rendered config (`/var/lib/kubelet/config.yaml`) once at startup and exposes systemReserved.cpu/systemReserved.memory as `neco_node_kubelet_system_reserved{node,resource,unit}`
  - the kubelet config is expected to be mounted at `/var/lib/kubelet/config.yaml`
- node label comes from a `NODE_NAME` env var, which is expected to be set from `spec.nodeName` in the DaemonSet manifest 
  - Setup() fails fast if `NODE_NAME` or either reserved value is missing/unparsable, since a silent fallback would make
downstream alerting meaningless.
- Value is read once and cached, no per-scrape file I/O, since this config only changes via a CKE-driven Node
restart anyway.


neco-exporter/e2e/kubeletconfig_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"strconv"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// findMetricValue scans Prometheus exposition text for a line starting with prefix
+// (a metric name plus its label set) and returns the trailing sample value.
+func findMetricValue(g Gomega, output []byte, prefix string) (float64, bool) {
```
yokaze: It looks better to move `findMetricValue` to `utils_test.go`?

yokaze (changes requested):
Almost LGTM; please rebase the branch to resolve conflict.

yokaze (approved):
LGTM

PR #2054, opened 2026-07-16
Update unbound root hints

terassyi:


unbound/build-targets.yaml
```diff
@@ -1,6 +1,7 @@
 - container-image: "unbound"
   make-targets: "check-generate"
   make-post-targets: "validate"
+  load: true
```
yokaze, line 4: Is it needed for some reason?

terassyi: Yes, `make validate` does `docker run` on the built image.
But with the buildx driver, the image isn't exported to the Docker's image store unless `load=true` is set.

yokaze (approved):
LGTM

PR #2058, opened 2026-07-17
fix trivy verification

pddg:
This includes two fixes.

## Support for cosign 3.0.6

In cosign 3.0.6, explicitly specifying the attestation type became mandatory as part of the response to GHSA-w6c6-c85g-mmv6. Therefore, --type slsaprovenance1 is now explicitly specified.

## Following the breaking changes in GitHub's attestation API

The verification procedure for the trivy binary provided by yamory relied on a deprecated parameter called .bundle. Following the introduction of the changes previously announced by GitHub, verification procedures depending on this parameter no longer work.
The response for .bundle_url, which was added as an alternative, is compressed with snappy, and there is no well-known effective method for decompressing this on Linux.

We considered using gh attestation verify, but found that it cannot be used for the trivy binary due to the following issue:
https://github.com/aquasecurity/trivy/discussions/10441

Upon examining the behavior of the gh CLI, we noticed that it uses the following library:
https://github.com/klauspost/compress
The GitHub Releases for this library provide a binary for compressing and decompressing the s2 format, which is compatible with snappy. By using this, it is possible to extract the bundle.json compressed in snappy format and perform verification with the cosign command as before.

### DIFF
--- a/.github/actions/setup-trivy/action.yaml
+++ b/.github/actions/setup-trivy/action.yaml
@@ -36,6 +36,7 @@ runs:
         cosign verify-blob-attestation \
           --bundle bundle.json \
           --new-bundle-format \
+          --type slsaprovenance1 \
           --certificate-identity "https://github.com/aquasecurity/trivy/.github/workflows/reusable-release.yaml@refs/tags/v${TRIVY_VERSION}" \
           --certificate-oidc-issuer 'https://token.actions.githubusercontent.com' \
           "${ARTIFACT}"
--- a/.github/actions/setup-trivy/action.yaml
+++ b/.github/actions/setup-trivy/action.yaml
@@ -6,7 +6,31 @@ inputs:
 runs:
   using: composite
   steps:
-    - env:
+    - name: Install s2d, snappy compatible unarchiver
+      env:
+        S2_VERSION: "v1.19.0"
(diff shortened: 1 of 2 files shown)


pddg:
@tkna @yokaze 
tcp-keepalive 0.3.6 is not scanned due to this problem.
https://github.com/cybozu/neco-containers/actions/runs/29554500036/job/87804851490

.github/actions/setup-trivy/action.yaml
```diff
@@ -6,7 +6,31 @@ inputs:
...
+        echo "::group::Download artifacts"
+        wget -O "${ARTIFACT}" "${S2_RELEASE_URL}/${S2_DEB}"
+        echo "::endgroup::"
+
+        echo "::group::Check checksum"
+        echo "${S2_CHECKSUM}  ${ARTIFACT}" | sha256sum -c -
+        echo "::endgroup::"
+
+        echo "::group::Install s2"
+        sudo apt-get install -y "${ARTIFACT}"
```
tkna, line 28: Don't we need `apt-get update` before install?

pddg: No need. The deb does not have any dependencies.

tkna (approved):
LGTM

PR #2060, opened 2026-07-21
Update bpftrace

terassyi:


yokaze (changes requested):
Could you please pin SHA256 of base images in `Dockerfile`?

yokaze (approved):
LGTM

PR #2061, opened 2026-07-21
neco-exporter: enable gofumpt and modernize

yokaze:
- **neco-exporter: use golangci-lint**
- **neco-exporter: enable gofumpt and modernize**
- **neco-exporter: bump version to 0.5.1**


tkna (approved):
LGTM

PR #2065, opened 2026-07-23
Revise linter setting

yokaze:
https://golangci-lint.run/docs/linters/false-positives/


tkna (approved):
LGTM

PR #2066, opened 2026-07-23
Regular update 2026-07 dap

kamito300:
part of https://bozuman.cybozu.com/k/63907/show#record=558


umezawatakeshi:
Why do you re-create patch files even if their "content" don't change ?
I think that those patches cannot be applied.

If patch files are updated even though their "content" don't change, it leads to the hassle of having to read patches that we don't need to read

umezawatakeshi (approved):

PR #2079, opened 2026-08-19
Add root golangci-lint configuration

ystkfujii:
This PR adds a golangci-lint configuration to the repository root.

It also adds an aqua configuration to manage the golangci-lint version.


tkna (approved):
LGTM

PR #2083, opened 2026-08-24
machines-endpoints: configure monitoring targets and sabakan address via flags, apply via SSA

masa213f:
Key changes:

- Stop discovering the boot server address via serf; require it to be specified explicitly with --sabakan-address (host:port, hostnames accepted).
- Replace --monitoring-endpoints with --boot-servers-port/--all-servers-port; each target's Service/EndpointSlice is only created when its port list is non-empty.
- Allow arbitrary ports (name:port, repeatable) on those targets instead of the previously hardcoded node-exporter/etcd-metrics ports, validated against Kubernetes' actual port-name constraints (max 15 characters, at least one letter, no leading/trailing/consecutive hyphens).
- Replace github.com/cybozu-go/log with the standard log/slog package.
- Service/EndpointSlice/ConfigMap writes now use server-side apply instead of a manual Get-then-Update-or-Create dance.
- Add --dry-run, which prints the JSON of the exact apply configuration that would be sent to Kubernetes instead of applying it.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


machines-endpoints/pkg/machines-endpoints/k8sresource.go
```diff
@@ -0,0 +1,195 @@
...
+		if slices.Contains(sliceNames, slice.Name) {
+			continue
+		}
+		err := endpointSliceInterface.Delete(ctx, slice.Name, metav1.DeleteOptions{})
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	// TODO remove transitive code
```
YZ775: In machines-endpoints v2, the name of the endpoints has changed to `all-servers-targets/boot-servers-targets`, but since the name in v1 is `prometheus-node-targets/http-node-exporter`, we'll need to use the old name.

machines-endpoints/pkg/machines-endpoints/k8sresource.go
```diff
@@ -0,0 +1,195 @@
...
+	} else if err == nil {
+		err := endpointInterface.Delete(ctx, name, metav1.DeleteOptions{})
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	// TODO remove transitive code
+	// Delete EndpointSlice with old name
+	_, err = endpointSliceInterface.Get(ctx, name, metav1.GetOptions{})
```
YZ775: Since this deletes the EndpointSlice based on `name`, it looks like `all-servers-targets/boot-servers-targets` will be deleted as well.

machines-endpoints/pkg/machines-endpoints/monitoring_endpoints.go
```diff
@@ -0,0 +1,107 @@
...
+			return nil, fmt.Errorf("invalid port %q (expected port:name)", spec)
+		}
+		port, err := strconv.ParseUint(portStr, 10, 16)
+		if err != nil {
+			return nil, fmt.Errorf("invalid port %q: %w", spec, err)
+		}
+		if port == 0 {
+			return nil, fmt.Errorf("invalid port %q: port must be between 1 and 65535", spec)
+		}
+		if errs := validation.IsValidPortName(name); len(errs) > 0 {
```
YZ775: Since `IsValidPortName` appears to be limited to 15 characters, using port names like `http-node-exporter` or `http-etcd-metric` will result in an error.

Since the actual validation of EndpointSlice ports seems to correspond to `IsDNS1123Label`, how about using that instead?

https://github.com/kubernetes/kubernetes/blob/fbb9a10c7a0469a6f076873289e18bd680402c80/pkg/apis/discovery/validation/validation.go#L182

machines-endpoints/machines-endpoints.yaml
```diff
@@ -103,7 +103,9 @@ spec:
               image: ghcr.io/cybozu/machines-endpoints:1.0.0
```
YZ775:
```suggestion
              image: ghcr.io/cybozu/machines-endpoints:2.0.0
```

machines-endpoints/README.md
```diff
@@ -22,16 +26,17 @@ Usage
...
+3. Check `all-servers-targets-0` EndpointSlice, `boot-servers-targets-0` EndpointSlice, `bmc-reverse-proxy` ConfigMap, and `bmc-log-collector` ConfigMap.
 
    ```console
-   kubectl get endpointslice -n NAMESPACE prometheus-node-targets-0
-   kubectl get endpointslice -n NAMESPACE bootserver-etcd-metrics-0
+   kubectl get endpointslice -n NAMESPACE all-servers-targets-0
+   kubectl get endpointslice -n NAMESPACE boot-servers-targets-0
    kubectl get configmap -n NAMESPACE bmc-reverse-proxy
+   kubectl get configmap -n NAMESPACE bmc-log-collector
    ```
```
YZ775: Could you add description of option like below ? (this sample is AI Generated)

| Option | Default | Description |
| ------ | ------- | ----------- |
| `--sabakan-address` | (required) | Address of sabakan's GraphQL API, in the form `host:port`. A hostname is accepted; prefer a name that resolves to every boot server, or a VIP, so that one unreachable boot server does not stop updates. |
| `--all-servers-port` | (none) | Port to expose on the `all-servers-targets` target, which lists all non-retired machines. In the form `port:name`, repeatable. The Service and EndpointSlices for the target are created only when at least one port is given. |
| `--boot-servers-port` | (none) | Same as `--all-servers-port`, but for the `boot-servers-targets` target, which lists non-retired boot servers only. |
| `--max-endpoints-per-slice` | `100` | Maximum number of endpoints per EndpointSlice. Must be between 1 and 1000. |
| `--bmc-reverse-proxy-configmap` | `false` | Generate the `bmc-reverse-proxy` ConfigMap. |
| `--bmc-log-collector-configmap` | `false` | Generate the `bmc-log-collector` ConfigMap. |
| `--dry-run` | `false` | Print the JSON of the resources that would be applied to stdout instead of applying them. A usable kubeconfig is still required, and EndpointSlices that would be deleted are not reported. |

YZ775 (changes requested):

machines-endpoints/pkg/machines-endpoints/monitoring_endpoints.go
masa213f: Thank you! Good!
I didn't knew the port name validation is differed depending on the resource...

https://github.com/kubernetes/kubernetes/blob/v1.36.4/pkg/apis/core/validation/validation.go#L2722

machines-endpoints/pkg/machines-endpoints/k8sresource.go
masa213f: Thank you. That's right.

However, in our actual cluster, the old Endpoints and EndpointSlices have already been deleted.
Since this transition logic shouldn't be necessary, I plan to remove it. Please wait while I check.

masa213f: This code was required for the upgrade from v0.9.11 to v1.0.0.
Since it is no longer needed in v2, I will remove it. 901b454

Also, I'll delete the v1's Service/EndpointSlice in the actual cluster manually.
So I don't implement migration code from v1 to v2.

machines-endpoints/pkg/machines-endpoints/k8sresource.go
masa213f: I'll delete it for the same reason as https://github.com/cybozu/neco-containers/pull/2083#discussion_r3867695767

machines-endpoints/pkg/machines-endpoints/monitoring_endpoints.go
masa213f: By the way, I also strengthened the validation of port numbers.

fd64979

YZ775 (approved):
LGTM

PR #2088, opened 2026-08-28
machines-endpoints: remove PodSecurityPolicy and hostNetwork

masa213f:
This PR clean up the sample manifest for machines-endpoints.
This does not affect the container image, so no tag update is needed.


YZ775 (approved):
LGTM

PR #2092, opened 2026-09-01
remove unused tools from golang images

pddg:


gokken-roko:
`TAG` files aren't updated. I think a new image will not be built or pushed after merging.
Is this intentional?

pddg:
Yes. This is intentional.
The next image releases for Go 1.26.8 and Go 1.27.1 are coming up soon, and it will be reflected in those releases.

gokken-roko (approved):
LGTM

PR #2095, opened 2026-09-02
machines-endpoints: Add label to service

masa213f:
This PR adds labels to the Service created by machines-endpoints.

When configuring targets in the VictoriaMetrics Operator, having labels on the Service makes it easier to select them (e.g. via ServiceScrape's selector).


zeroalphat (approved):
LGTM

PR #2098, opened 2026-09-02
bmc-log-collector: collect the iDRAC Lifecycle log

shunki-fujita:
## What

Collect the LC (Lifecycle) log from iDRAC in addition to the SEL, so that it can be persisted in Loki.

- The LC log is collected in the same way as the SEL: one request per cycle, the entries newer than the pointer are emitted, and the pointer is advanced to the newest Id. The LC log endpoint returns only the latest 50 entries, so the entries that fell off the page since the previous cycle are not collected (a warning is logged). This is accepted because the LC log grows only ~2.3 entries/day/machine.
- The first collection for a machine and the collection after the LC log was cleared in iDRAC emit the latest page, so that the whole history is not ingested at once. The clear is detected by a newest Id smaller than the pointer, or by a different creation time of the pointered Id.
- The pointer is advanced as in the SEL collection: an unparsable Id, an unparsable creation time used for the clear detection, or a failed write aborts the cycle and keeps the pointer unchanged; an entry that cannot be marshaled is skipped.
- A BMC that replies 404/405 for the LC log path is treated as "LC log service not implemented" and is not counted as a failure, to avoid a permanent false alarm.
- Each output line now carries a `LogType` field (`SEL` or `LCLog`). This is an additive field, so the existing Loki queries keep working.
- The request counters have a new `log_type` label (`sel` / `lclog`). The existing `FailedToConnectBMC` alert rule aggregates with `sum by(serial)`, so it is compatible as is.
- The pointer file gets new `Lc*` fields. Files written by older versions are loaded with zero values, so the LC log starts fresh without migration.
- SEL and LC log are collected sequentially in the same goroutine to avoid concurrent accesses to the same iDRAC.

## Design notes

The endpoint behavior (newest-first order, numeric contiguous Ids, latest 50 entries per request) and the log flow rate (~2.3 entries/day/machine) were verified on stage0 machines (iDRAC FW 7.20.30.55) before the implementation. See `docs/design.md` for details.

## How to test

- `make test` (ginkgo with `-race`): new specs cover the first collection, new entries, no-change cycles, log clear (both detection paths), more new entries than the page holds, malformed entries, an empty LC log, an unparsable creation time, a write failure, an LC-less BMC, and old pointer file compatibility.


bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,248 @@
...
+			if id == lastPtr.LcLastReadId && lastPtr.LcLastReadCreateTime != 0 {
+				createTime, err := time.Parse(time.RFC3339, v.Create)
+				if err == nil && createTime.Unix() != lastPtr.LcLastReadCreateTime {
+					// The same Id with a different creation time: the LC log was
+					// cleared and has grown beyond the last read Id since then
+					cleared = true
+					break scan
+				}
+			}
+			foundKnown = true
```
ystkfujii: I'm not sure `foundKnown` exactly represents the case where we want to increment counterLcCatchupTruncated.
For example, `foundKnown` is still false when the loop stops because an empty page is returned.

Would it be simpler and more direct to check whether the loop actually reached lcMaxPages?

WDYT?

```
page := 0
for ; page < c.lcMaxPages; page++ {
    ...
}

if len(newLogs) > 0 && page == c.lcMaxPages {
    ...
}
```

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,248 @@
...
+				cleared = true
+				break
+			}
+		}
+
+		for _, v := range response.Members {
+			id, err := strconv.Atoi(v.Id)
+			if err != nil {
+				slog.Error("failed to strconv", "err", err, "serial", m.Serial, "Id", v.Id)
+				continue
```
ystkfujii: How should we handle an invalid LC log entry?

The error handling seems a little inconsistent. 
Some parsing errors stop the collection, while others skip the entry or ignore the error.

Since `v.Id` and `v.Create` are used for pointer management and log-clear detection, continuing after these errors may cause entries to be skipped permanently.

Would it be safer to stop the current collection and leave the pointer unchanged when these values cannot be parsed?

bmc-log-collector/README.md
```diff
@@ -45,9 +47,10 @@ bmc-log-collector command provides the usage in following.
 $ bmc-log-collector --help
 
 Usage of ./bmc-log-collector:
-      --bmc-user-json string          User and password of BMC (default "/users/neco/bmc-user.json")
-      --machine-list-json string      Target machines list of log scraping (default "/config/machineslist.json")
-      --pointer-dir-path string       Data directory of pointer management (default "/data/pointers")
-      --scraping-interval-timer int   Timer(sec) of scraping interval time (default 300)
-      --user-id string                User ID of bmc-user-json JSON file (default "support")
+      --bmc-user-json string         User and password of BMC (default "/users/neco/bmc-user.json")
+      --lclog-max-pages int          Maximum pages of the lifecycle log to read per scraping cycle (default 40)
```
ystkfujii: > The paging behavior (newest-first order, numeric contiguous Ids, $skip) and the log flow rate (~2.3 entries/day/machine) were verified on stage0 machines (iDRAC FW 7.20.30.55) before the implementation. See docs/design.md for details.

At about 2.3 entries/day/machine, 40 pages (2,000 entries) covers roughly 870 days of logs.

Do we need such a large default? I'm a little concerned about the load on the BMC. Could we use a smaller default, such as 3 pages, and increase it only when necessary?

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,248 @@
...
+			lastPtr.LcLastHttpStatusCode = 0
+			lastPtr.LcLastError = err.Error()
+			if err := updateLastPointer(lastPtr, filePath); err != nil {
+				slog.Error("failed to write a pointer file.", "err", err, "serial", m.Serial, "filePath", filePath)
+			}
+			return
+		}
+		if statusCode == http.StatusNotFound || statusCode == http.StatusMethodNotAllowed {
+			// The BMC does not implement the LC log service.
+			// Not counted as a failure to avoid a permanent false alarm.
```
ystkfujii: Do we need to check for an unsupported LC service on every page?

If page 0 returns 200, we already know that the LC log service is supported. A 404 or 405 on a later $skip request seems like a different error and probably should not be treated as "LC log service not implemented."

---
It might also be useful to expose whether LC logging is supported as a metric. This could be done in a separate PR.

bmc-log-collector/testdata/redfish_response/LCLOG01-lc-1.json
ystkfujii, line 1: How were these Redfish response fixtures created?

Are they sanitized responses from a real iDRAC, or manually created test data?

Since this logic depends on the actual response structure, ordering, and ID behavior, I'd like to make sure the fixtures reflect the behavior of the real devices.

---
Also, do the actual iDRAC responses include `Members@odata.nextLink` when more entries are available? The mock currently paginates the snapshots using `$skip`, but does not include `Members@odata.nextLink`.

https://redfish.dmtf.org/schemas/DSP0266_1.20.1.html#next-link-membersodatanextlink-property

bmc-log-collector/metrics.go
```diff
@@ -9,18 +9,32 @@ import (
...
+	[]string{"serial", "log_type"},
 )
 
 var counterRequestSuccess = promauto.NewCounterVec(
 	prometheus.CounterOpts{
 		Name: "bmc_log_requests_success_total",
-		Help: "Succeeded count of accessing BMC to get the system event log",
+		Help: "Succeeded count of accessing BMC to get the hardware log",
+	},
+	[]string{"serial", "log_type"},
```
ystkfujii, line 31: > The request counters have a new log_type label (sel / lclog). The existing FailedToConnectBMC alert rule aggregates with sum by(serial), so it is compatible as is.

I think we should also confirm that the meaning of the existing request counters and alert thresholds remains appropriate.

Previously, each scraping cycle made one SEL request per BMC. With this change, an unreachable BMC can increment the failed counter for both SEL and LC log requests, resulting in +2 per cycle.

The success counter can also increase multiple times in a single cycle when LC log catch-up requires multiple pages.

`sum by(serial)` keeps the existing PromQL working, but the rate at which these counters increase changes, so I think it would be worth checking the existing alert thresholds.

bmc-log-collector/main.go
```diff
@@ -97,6 +100,7 @@ var (
 	flgMachineList          *string = pflag.String("machine-list-json", "/config/machineslist.json", "Target machines list of log scraping")
 	flgPointerDir           *string = pflag.String("pointer-dir-path", "/data/pointers", "Data directory of pointer management")
 	flgScrapingIntervalTime *int    = pflag.Int("scraping-interval-time", 300, "Timer(sec) of scraping interval time")
+	flgLcMaxPages           *int    = pflag.Int("lclog-max-pages", 40, "Maximum pages of the lifecycle log to read per scraping cycle")
```
ystkfujii: > The paging behavior (newest-first order, numeric contiguous Ids, $skip) and the log flow rate (~2.3 entries/day/machine) were verified on stage0 machines (iDRAC FW 7.20.30.55) before the implementation. See docs/design.md for details.

At about `2.3 entries/day/machine`, 40 pages (2,000 entries) would cover roughly 870 days of logs.

Since `--lclog-max-pages` is only an upper bound and normally we should stop as soon as the previous pointer is found, this may not be an issue in practice. However, in the worst case it could result in up to 40 requests to a BMC in one collection cycle.

Is there a particular recovery period or scenario that the default of 40 pages is intended to cover? I wonder if a smaller default, such as 3 pages, might be sufficient.

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,248 @@
...
+	var newestCreateTime int64
+	foundKnown := false
+	cleared := false
+	skip := 0
+
+scan:
+	for page := range c.lcMaxPages {
+		bmcUrl := baseUrl
+		if skip > 0 {
+			bmcUrl = baseUrl + "?$skip=" + strconv.Itoa(skip)
```
ystkfujii: https://github.com/cybozu/neco-containers/pull/2098/changes#r3923976078

If `Members@odata.nextLink ` is available, I think it would be better to use it instead of constructing the URL manually like this:

```
bmcUrl = baseUrl + "?$skip=" + strconv.Itoa(skip)
```

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+			if id > lastPtr.LcLastReadId {
+				// An entry created between the page requests shifts the $skip
+				// offset backward and the next page repeats the entries of the
+				// previous page. Skip the already collected Ids.
+				if len(newLogs) == 0 || id < minCollectedId {
+					newLogs = append(newLogs, v)
+					minCollectedId = id
+				}
+				continue
```
ystkfujii: The `minCollectedId` logic is a little hard to follow.

Since there are at most around 2,000 entries, I think using a map to track seen IDs would be easier to understand, even though it uses a bit more memory.

```
seen := map[string]struct{}{}

for _, v := range response.Members {
    if _, ok := seen[v.Id]; ok {
        continue
    }
    seen[v.Id] = struct{}{}

    ...
}
```


Also, depending on the default value, it could end up being quite low.

- Ref: https://github.com/cybozu/neco-containers/pull/2098#discussion_r3924038687

bmc-log-collector/lclog-collector.go
shunki-fujita: Good point — fixed in 818ab538. The metric is now counted only when the loop actually reaches `lcMaxPages`; reaching the end of the log without finding the pointered entry is reported with a separate warning instead (it suggests an undetected log clear). Added a test for the latter case.

bmc-log-collector/lclog-collector.go
shunki-fujita: Agreed. Fixed in 818ab538: parse failures of `Id`/`Created` now abort the cycle without advancing the pointer, so the next cycle retries (test added). Also applied the same policy to the SEL collector in 7a3a0a1e.

bmc-log-collector/lclog-collector.go
shunki-fujita: Fixed in 818ab538: 404/405 is treated as "not implemented" only on the first page; the same status on later pages is handled as an ordinary failure. +1 on exposing LC support as a metric in a separate PR.

bmc-log-collector/testdata/redfish_response/LCLOG01-lc-1.json
shunki-fujita: Hand-made, but the field structure is copied from a response captured on a real iDRAC (stage0, FW 7.20.30.55), and the newest-first ordering / contiguous numeric Ids were verified on the same device beforehand. I also confirmed on the real device that `Members@odata.nextLink` IS returned while more entries are available and omitted on the last page — the implementation now follows nextLink (818ab538) and the mock reproduces this behavior.

bmc-log-collector/metrics.go
shunki-fujita: Good catch. An unreachable BMC now increments the failed counter at 24/h instead of 12/h, so detection still works but sensitivity to transient failures roughly doubles (2 failed cycles now cross the >3 threshold instead of 4). I'll raise the threshold to 6 in the neco-apps PR that bumps the image tag. The success counter is not referenced by any rule.

bmc-log-collector/main.go
shunki-fujita: The limit is a safety bound for a collector outage combined with maintenance bursts (FW updates write many LC entries at once), not the normal working set — a regular cycle stops at the first page as soon as the pointer is found. 3 pages (150 entries) could truncate after such a case, so I lowered the default to 10 pages (500 entries) as a middle ground in 9cd96f58.

bmc-log-collector/lclog-collector.go
shunki-fujita: Done in 818ab538 — verified on a real iDRAC that nextLink is returned while more entries are available and omitted on the last page (see also the fixture thread).

bmc-log-collector/lclog-collector.go
shunki-fujita: Agreed, replaced with a seen map in 818ab538. (For the record, the initial-value concern was guarded by `len(newLogs) == 0`, but needing that explanation proves your readability point.)

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,263 @@
...
+		v.LogType = logTypeLCLog
+
+		bmcByteJsonLog, err := json.Marshal(v)
+		if err != nil {
+			slog.Error("failed to marshal the lifecycle log", "err", err, "serial", m.Serial, "Id", v.Id)
+			continue
+		}
+		err = logWriter.write(string(bmcByteJsonLog), m.Serial)
+		if err != nil {
+			slog.Error("failed to output log", "err", err, "serial", m.Serial, "bmcByteJsonLog", string(bmcByteJsonLog))
```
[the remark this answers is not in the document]
shunki-fujita: Fair point — the production writer (`log.Print` to stdout) cannot actually return an error, but the at-least-once guarantee should not depend on that. Fixed in 03582261: a write failure now stops the emission and the pointer is advanced only when all the entries were written, so the next cycle re-emits them. Applied to both the LC and the SEL collectors, with a test.

bmc-log-collector/main.go
```diff
@@ -97,6 +100,7 @@ var (
 	flgMachineList          *string = pflag.String("machine-list-json", "/config/machineslist.json", "Target machines list of log scraping")
 	flgPointerDir           *string = pflag.String("pointer-dir-path", "/data/pointers", "Data directory of pointer management")
 	flgScrapingIntervalTime *int    = pflag.Int("scraping-interval-time", 300, "Timer(sec) of scraping interval time")
+	flgLcMaxPages           *int    = pflag.Int("lclog-max-pages", 10, "Maximum pages of the lifecycle log to read per scraping cycle")
```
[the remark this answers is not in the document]
shunki-fujita: The default was intentionally lowered from 40 to 10 during the review (see the discussion above); the PR description was stale. Updated the description to match the implementation (default 10, nextLink-based paging).

bmc-log-collector/log-collector.go
```diff
@@ -127,23 +136,31 @@ func (c *selCollector) collectSystemEventLog(ctx context.Context, m Machine, log
 	for i, v := range slices.Backward(response.Sel) {
 		currentId, err := strconv.Atoi(v.Id)
 		if err != nil {
-			slog.Error("failed to strconv", "err", err, "serial", m.Serial, "LastReadId", currentId, "ptrDir", c.ptrDir)
-			continue
+			// The Id is the basis of the pointer management. When it cannot be
+			// parsed, abort this cycle without updating the pointer file so that
+			// no entry is skipped permanently; the next cycle retries.
+			slog.Error("failed to strconv; abort this cycle to keep the pointer unchanged", "err", err, "serial", m.Serial, "Id", v.Id, "ptrDir", c.ptrDir)
+			return
```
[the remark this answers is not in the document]
shunki-fujita: Agreed — the SEL collector emits while iterating, so the abort could re-emit the earlier entries every cycle (the test output showed the duplicated Id 3). Fixed in 61b270e7: all the Ids are now validated before any entry is written, so a persistently malformed entry produces only an error log per cycle, with no duplicates and the pointer unchanged. The LC collector already buffers and emits only after a fully validated scan, so it did not need this change. Note that the remaining blockage while the malformed entry persists is intentional: silently skipping it would lose the entry permanently, and the repeated error log makes the condition visible.

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,271 @@
...
+
+			// The first collection for this machine: start from the latest page without backfill
+			if lastPtr.LcLastReadId == 0 {
+				break
+			}
+			// The entry Id restarts from 1 when the LC log is cleared in iDRAC
+			if newestId < lastPtr.LcLastReadId {
+				cleared = true
+				break
+			}
```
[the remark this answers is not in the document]
shunki-fujita: Right, the early exits bypassed the per-entry validation. Fixed in the latest commit: every member of page0 is now validated before the initial-collection / after-clear exits, so a malformed entry aborts the cycle with the pointer unchanged, same as the catch-up path. Added a test (malformed non-first entry on the initial collection, recovered in the next cycle).

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,262 @@
...
+
+// collectLifecycleLog collects the LC (Lifecycle) log from iDRAC.
+//
+// Unlike the SEL endpoint, the LC log endpoint returns only the latest page
+// (50 entries on the real iDRAC, newest first). This function follows
+// Members@odata.nextLink backward until it finds the entry read in the
+// previous cycle, up to lcMaxPages pages. The first collection for a machine
+// and the collection after the LC log was cleared in iDRAC go through the
+// same loop; the page limit bounds the backfill in those cases.
+func (c *logCollector) collectLifecycleLog(ctx context.Context, m Machine, logWriter bmcLogWriter) {
```
ystkfujii, line 70: What do you think about introducing types like the following?

```go
type lcScanStopReason int

const (
	lcScanStopReasonFoundPointer lcScanStopReason = iota
	lcScanStopReasonInitial
	lcScanStopReasonCleared
	lcScanStopReasonExhausted
	lcScanStopReasonPageLimit
)

type lcScanResult struct {
	logs             []LifeCycleLog
	newestID         int
	newestCreateTime int64
	stopReason           lcScanStopReason
}
```

I think readability could also be improved by moving the logic for collecting LifeCycleLog entries into a separate method, for example:

```go
result, ok := c.scanLifecycleLogs(ctx, m, &lastPtr, filePath)
  ...

switch result.reason {
case lcScanCleared:
   ...
case lcScanPageLimit:
   ...
case lcScanExhausted:
   ...
}
```

This would make the reason why the scan stopped more explicit and simplify the control flow in collectLifecycleLog.

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		slog.Error("can't check a pointer file.", "err", err, "serial", m.Serial, "filePath", filePath)
+		return
+	}
+
+	lastPtr, err := readLastPointer(filePath)
+	if err != nil {
+		slog.Error("can't read a pointer file.", "err", err, "serial", m.Serial, "filePath", filePath)
+		return
+	}
```
ystkfujii, line 77: I think this can be made DRY.

https://github.com/cybozu/neco-containers/blob/main/bmc-log-collector/log-collector.go#L60-L70

bmc-log-collector/lclog-collector.go
ystkfujii: Removing the flag-based state management made the code much easier to read.

Looking at it again, the previous state management was mainly used for logging and metrics, so it may have been a bit too complex for that purpose. I think we could also remove `lcScanResult` and handle the necessary logic directly inside scanLifecycleLog. That may make the code even simpler.

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		slog.Error("can't load a pointer file.", "err", err, "serial", m.Serial, "filePath", filePath)
+		return
+	}
+
+	result, ok := c.scanLifecycleLog(ctx, m, &lastPtr)
+	if !ok {
+		// The failure has been reported; record the request status and keep
+		// the read position unchanged so that the next cycle retries
+		if err := updateLastPointer(lastPtr, filePath); err != nil {
```
ystkfujii: Go has `defer`, so I think we may be able to use it to reduce some duplicated code.

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,305 @@
...
+	for page := 0; page < c.lcMaxPages; page++ {
+		response, ok := c.fetchLifecycleLogPage(ctx, m, lastPtr, url, page == 0)
+		if !ok {
+			return lcScanResult{}, false
+		}
+		if len(response.Members) == 0 {
+			if page == 0 {
+				result.stop = lcScanEmptyLog
+			} else {
+				result.stop = lcScanEndOfLog
```
ystkfujii: When the same variable is updated in multiple places, the reader has to follow the whole function to understand its final value.

If possible, I think it would be easier to read if we create the needed value at that point and return it directly. This way, the reader does not need to keep tracking how the value may change later.

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,305 @@
...
+
+	// Not finding the last read entry is an anomaly only during a catch-up;
+	// stopping at the page limit while collecting from scratch just bounds
+	// the backfill. Reported only after the new position is persisted: when
+	// the emission or the pointer write above fails, the next cycle retries
+	// from the old position and no entry is skipped
+	if result.catchingUp() {
+		switch result.stop {
+		case lcScanPageLimit:
+			counterLcCatchupTruncated.WithLabelValues(m.Serial).Inc()
```
ystkfujii: This may be a matter of preference, but I think the metric wording could be a little simpler.

Rather than describing it as the number of times the result was truncated, I think describing it as the number of times the page limit was reached would more directly represent what actually happened.

Also, using the existing term page limit instead of introducing a new concept like truncated may make the code easier to understand.

bmc-log-collector/lclog-collector.go
```diff
@@ -0,0 +1,305 @@
...
+	newestId         int            // Id of the newest entry; 0 when the log is empty
+	newestCreateTime int64          // Created time of the newest entry
+	lastReadId       int            // the Id the scan searched for; 0 when collecting from scratch
+	stop             lcScanStop
+}
+
+// catchingUp reports whether the scan searched for the entry read in the
+// previous cycle, as opposed to collecting from scratch (the first collection
+// for a machine, or the collection after the LC log was cleared in iDRAC).
+func (r lcScanResult) catchingUp() bool {
```
ystkfujii: This may be a matter of preference, but I personally find introducing a target struct easier to understand.

bmc-log-collector/lclog-collector.go
shunki-fujita: Done in bc0ceacc: dropped `lcScanStop` and the `switch`; the warnings are now emitted inside `scanLifecycleLog` where the outcome is known. The result keeps only the entries, the newest position, and `gap` (the scan hit the page limit before reaching the target). `gap` stays in the result because the page limit metric/warning must be reported only after the pointer write succeeds (the ordering fixed for the Copilot comment); counting inside the scan would count a retried cycle twice.

I also looked at the wip branch, thanks. I kept four behaviors it changes since they were agreed earlier: backfill up to the page limit on the first collection / after a clear (design.md), validation of every page0 member before emitting, 404/405 as "not implemented" only on the first LC page (SEL 404 still counts as a failure), and the deduplication of repeated failure logs (without it, dctest emits 16 WARN lines every cycle).

bmc-log-collector/lclog-collector.go
shunki-fujita: I tried `defer`, but the page limit metric/warning has to come after a successful pointer write, and a deferred write runs last. So the write stays explicit; I wrapped it and its error report in `saveLastPointer`, which also replaced the duplicated blocks in the SEL collector.

bmc-log-collector/lclog-collector.go
shunki-fujita: `result.stop` is gone. The `lastReadId` reset is now `target = nil` at the two clear-detection points; the append to `result.logs` and the `newestId` assignment remain since they accumulate the collected data.

bmc-log-collector/lclog-collector.go
shunki-fujita: Renamed to `bmc_lclog_page_limit_reached_total` and replaced `catchingUp()` with a direct `target != nil` check. design.md is updated accordingly.

bmc-log-collector/lclog-collector.go
shunki-fujita: Introduced `lcScanTarget` (Id and Created of the entry read in the previous cycle); it is nil on the first collection and after a clear.

bmc-log-collector/metrics.go
```diff
@@ -9,18 +9,32 @@ import (
...
+)
+
+var counterLcPageLimitReached = promauto.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "bmc_lclog_page_limit_reached_total",
```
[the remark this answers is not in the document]
shunki-fujita: The PR description was stale. The metric was renamed to `bmc_lclog_page_limit_reached_total` on the reviewer's request (see the thread on the metric wording). It is new in this PR and nothing consumes it yet, so there is no existing dashboard or alert to break. Updated the description to match the implementation and design.md.

bmc-log-collector/docs/design.md
```diff
@@ -42,6 +42,54 @@ flowchart TB
...
+
+1. Use `/redfish/v1/Managers/iDRAC.Embedded.1/LogServices/Lclog/Entries` as the path to Redfish.
+2. This endpoint returns only the latest 50 entries in the newest-first order.
+   The collector follows `Members@odata.nextLink` backward until it finds the entry
+   whose ID was recorded in the pointer file in the previous cycle. On the real iDRAC
+   (verified on FW 7.20.30.55), `Members@odata.nextLink` is returned while more entries
+   are available and is omitted on the last page.
+   The number of pages read in one cycle is limited (3 pages by default). When the limit is hit,
+   the collector emits only the entries it has read, records the gap in the
+   `bmc_lclog_page_limit_reached_total` metric, and continues from the newest entry.
```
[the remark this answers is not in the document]
shunki-fujita: The PR description was stale. The metric was renamed to `bmc_lclog_page_limit_reached_total` on the reviewer's request (see the thread on the metric wording). It is new in this PR and nothing consumes it yet, so there is no existing dashboard or alert to break. Updated the description to match the implementation and design.md.

bmc-log-collector/request-to-bmc.go
```diff
@@ -27,3 +29,48 @@ func requestToBmc(ctx context.Context, username string, password string, client
...
+		}
+		*lastHttpStatusCode = statusCode
+		*lastError = ""
+		return nil, false
+	}
+	counterRequestSuccess.WithLabelValues(m.Serial, logType).Inc()
+	// Clear the last error status so that the same error after a recovery
+	// is logged again instead of being suppressed by the deduplication
+	*lastHttpStatusCode = statusCode
+	*lastError = ""
```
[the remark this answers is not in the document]
shunki-fujita: Right, a success on the first page erased the failure recorded for a later page. Fixed in 5595f496: `requestBmcLog` now records failures only, and the collectors clear the status once the whole collection succeeded, so a consistently failing later page is reported once.

bmc-log-collector/lclog-collector.go
shunki-fujita: Update: in 1ebf56f3 the first collection and the collection after a clear now read only the latest page (back to the original design; a single request decides the outcome, which suits unreliable BMCs). The scan no longer resets its target mid-loop, and reaching the page limit always means a gap, which removes the remaining in-loop state updates. design.md, README, and the PR description are updated.

ystkfujii (changes requested):
**1. Pointer update**

* Following up on my previous comment, the parsing of `Created` still seems different from `SystemEventLog`.
* Currently, if parsing fails, `0` is written to the pointer. I think this may cause problems on the next read.

**2. The `The first collection for the machine` branch seems unnecessary**

My understanding is:

* `SystemEventLog` does not have this special case.
* Even if we keep it, `LcLastReadCreateTime` is `0` on the first collection anyway, so this branch seems unnecessary.

**3. Is sorting by ID necessary?**

* `SystemEventLog` handles the order with `slices.Backward`.
* I would understand sorting if the API could return entries in arbitrary order, but the PR says that newest-first ordering was verified, so I am a little confused about why sorting is needed.
* If arbitrary ordering can actually happen, I think we should handle it for both `SystemEventLog` and `LifecycleLog`, preferably in a separate PR.

**4. Remove unnecessary comments**

Comments also have a cost, so please remove comments that do not add useful information.

* Reading cost: every comment adds cognitive load, and too many comments can make the implementation harder to understand.
* Maintenance cost: comments need to stay consistent with the code when the implementation changes.

There are different opinions on what makes a comment useful, but t_wada once said that comments should explain things that were intentionally *not* done. I think that is a good guideline here, so it would be helpful to remove comments that simply describe what the code already says.

GoDoc is a little different, but I think some of it could also be shorter.

For example:

* `All the entries up to the newest one were written`: this just describes the following code and is easy to understand from the code itself.
* `The entries between the pointer ...`: the following log message already explains the same thing, so this seems redundant.

This topic is also covered in [The Art of Readable Code](https://www.oreilly.co.jp/books/9784873115658/) if you are interested.

**5. Question: Is there any difference between the `SystemEventLog` and `LifecycleLog` APIs?**

My expectation is that we should first try to collect `LifecycleLog` using the same logic as `SystemEventLog`.

However, the current implementation uses different logic in several places, and I am wondering why these differences are necessary.

I have not been able to test this against a real iDRAC myself, so I may be missing some API behavior or implementation detail.

PR #2101, opened 2026-09-03
Introduce pinact

takahiro-yamada:
Introduced [pinact](https://github.com/suzuki-shunsuke/pinact) to manage and verify pinned GitHub Actions versions, and extracted aqua setup into a shared composite action.

### DIFF
--- a/.github/actions/aqua/action.yaml
+++ b/.github/actions/aqua/action.yaml
@@ -1,0 +1,11 @@
+name: "Setup tools"
+description: "Setup tools with aqua"
+runs:
+  using: composite
+  steps:
+    - uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e # v4.0.5
+      with:
+        aqua_version: v2.62.3
+        aqua_opts: ""
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
--- a/.github/actions/aqua/action.yaml
+++ b/.github/actions/aqua/action.yaml
@@ -6,6 +6,5 @@ runs:
     - uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e # v4.0.5
       with:
         aqua_version: v2.62.3
-        aqua_opts: ""
       env:
         GITHUB_TOKEN: ${{ github.token }}
--- a/.github/actions/aqua/action.yaml
+++ b/.github/actions/aqua/action.yaml
@@ -6,5 +6,3 @@ runs:
     - uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e # v4.0.5
       with:
         aqua_version: v2.62.3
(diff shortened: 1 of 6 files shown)


.github/actions/aqua/action.yaml
```diff
@@ -0,0 +1,11 @@
+name: "Setup tools"
+description: "Setup tools with aqua"
+runs:
+  using: composite
+  steps:
+    - uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e # v4.0.5
+      with:
+        aqua_version: v2.62.3
+        aqua_opts: ""
```
ystkfujii: Is `aqua_opts: ""` necessary here?

Without it, aqua-installer uses `-l` by default, so tools like pinact are installed lazily only when needed.

https://github.com/aquaproj/aqua-installer/blob/96a9bc20066c5bf5e275b41019cfc165b25f4e2e/action.yaml#L17

```suggestion
```

.github/workflows/pinact-verify.yaml
```diff
@@ -0,0 +1,22 @@
...
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/aqua
+      - name: Verify SHAs and tag comments are consistent
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: pinact run --check --verify-comment
```
ystkfujii: Should we also enforce the 14-day minimum age in CI?

```suggestion
        run: pinact run --check --verify-comment --min-age 14
```

ystkfujii (changes requested):

.github/actions/aqua/action.yaml
takahiro-yamada: I think it's necessary when leveraging caching, but since we don't use cache here and only install pinact via aqua in CI, it's not needed. I'll remove it, thanks.

.github/workflows/pinact-verify.yaml
takahiro-yamada: I agree, fixed: [0d86254](https://github.com/cybozu/neco-containers/pull/2101/commits/0d862540792daaf56733f9a2c932d944721f4866)

.github/actions/aqua/action.yaml
takahiro-yamada: fixed: [44dd651](https://github.com/cybozu/neco-containers/pull/2101/commits/44dd651ad5fb4e2e05bb3186c64822fe464e5c80)

Besides, I removed a redundant specification of github_token in [3c00e82](https://github.com/cybozu/neco-containers/pull/2101/commits/3c00e8200410e4d88cdc3801f62d4005e49a9fb5).

ystkfujii (approved):

PR #2102, opened 2026-09-03
upgrade pomerium 0.33.1 and use distroless instead of ubuntu

pddg:
In daily vulnerability management, handling internet-facing components like Pomerium requires extra care. However, numerous vulnerabilities in the ubuntu image are reported despite having no actual impact.

Envoy uses BoringSSL rather than OpenSSL. Additionally, Pomerium and its embedded Envoy are statically built, relying on no shared libraries in the container image. Upstream also uses `distroless/base-nossl` for release images, so switching from ubuntu offers great value by reducing vulnerability management overhead.

While `distroless/static` would likely work, we will align with upstream and use `base-nossl`.


takahiro-yamada (approved):
LGTM

PR #2109, opened 2026-09-11
hubble-ui, argocd: set pnpm minimumReleaseAge to 3 days

shunki-fujita:
## What

Set the pnpm waiting period (`minimumReleaseAge`) to 3 days (4320 minutes) wherever pnpm resolves versions in this repository:

- `hubble-ui/Makefile`: pass `--config.minimumReleaseAge=$(PNPM_MINIMUM_RELEASE_AGE)` (default 4320) to the `pnpm install` in `import-lockfile`. Until now this relied on the pnpm 11 default of 1 day. Upstream ships its own `.npmrc`, so the value is passed as a flag instead of copying a config file into the checkout.
- `argocd/Dockerfile`: raise `pnpm config set -g minimumReleaseAge` from 1440 to 4320. This build installs with `--frozen-lockfile`, so the setting is only consulted if pnpm ever has to resolve; the change is for consistency and to avoid suggesting a 1-day policy.
- `maintenance.md`: note the 3-day rule in the hubble-ui lockfile procedure, and how to override it once (`PNPM_MINIMUM_RELEASE_AGE=0`) when upstream genuinely needs a version younger than 3 days.

## Why

The company-wide npm / pnpm security guideline recommends `minimumReleaseAge: 4320` (3 days) for pnpm projects so that freshly published versions are not picked up before they have had time to be inspected or pulled. The hubble-ui lockfile import is the one place in this repository where pnpm resolves new versions, and it was using the 1-day default.

## Not changed

- grafana (Yarn 4, `--immutable`) and pomerium (`npm ci`) install from upstream lockfiles and never resolve versions, so no waiting-period setting applies to them.
- No `TAG` is bumped: neither change alters image contents. The argocd value takes effect on its next regular rebuild.

## Verification

- pnpm versions in use (hubble-ui 11.1.3 via aqua, argocd 10.34.5) are above the guideline's minimum of 10.26.0 and both support `minimumReleaseAge` (added in 10.16.0).
- `make -C hubble-ui import-lockfile` was not re-run in this PR; the lockfile is unchanged. The flag only affects future lockfile regenerations.

### DIFF
--- a/argocd/Dockerfile
+++ b/argocd/Dockerfile
@@ -76,7 +76,9 @@ ENV PATH="/opt/node-v${NODE_VERSION}-linux-x64/bin:${PATH}"
 
 RUN --mount=type=bind,from=pnpm-download,source=/pnpm,target=/tmp/pnpm \
     install -m 0755 /tmp/pnpm /usr/local/bin/pnpm && \
-    pnpm config set -g minimumReleaseAge 1440
+    # 3 days, per the company-wide npm / pnpm guideline. Only used if pnpm has to
+    # resolve versions; the frozen-lockfile install below does not resolve anything.
+    pnpm config set -g minimumReleaseAge 4320
 
 RUN apt-get update && apt-get install -y --no-install-recommends gpg && \
     # Unlike upstream, do not remove /usr/share/doc/* to preserve license files
--- a/hubble-ui/Makefile
+++ b/hubble-ui/Makefile
@@ -5,6 +5,9 @@ NGINX_COMMIT_HASH ?= c743d5145c69250f030934280b01b638d9d765ed
 PNPM_DIR ?= $(CURDIR)/.pnpm-bin
 PNPM_BIN ?= $(CURDIR)/.pnpm-bin/pnpm
 TAKUMI_GUARD_REGISTRY ?= https://npm.flatt.tech/
+# Do not resolve versions published less than 3 days (4320 minutes) ago.
+# Matches the company-wide npm / pnpm security guideline.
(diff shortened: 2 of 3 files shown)


takahiro-yamada (changes requested):
Thanks for PR, but I’d prefer not to adopt this setting for the Argo CD build image for the following reasons:

- Argo CD is not built frequently.
- Builds rely on the lockfile once updated, so the ongoing risk is low.
- In situations where an urgent security patch is required, this restriction could block immediate updates.

Given these points, the risk of delaying critical security fixes outweighs the benefits here.

Cc: @cybozu/pdx

shunki-fujita:
@takahiro-yamada 
 - This setting has no effect on the current build. pnpm install --frozen-lockfile never resolves versions, and pnpm only applies minimumReleaseAge while resolving. Neither the existing 1440 nor 4320 can delay an urgent ARGOCD_VERSION bump.
 - It would only kick in if a resolving step is added later (dropping --frozen-lockfile, pnpm add/update). That is where a 3-day default protects against a freshly published compromised version, and it can be bypassed per command (--config.minimumReleaseAge=0) or per package (minimumReleaseAgeExclude) when an urgent fix is needed.
  
So I'd like to keep 4320 as a default that only matters if the build ever starts resolving.

takahiro-yamada (approved):
@shunki-fujita 
Thanks for the explanation. Makes sense that `--frozen-lockfile` won't be blocked by this during emergency version bumps, and that this serves as a safeguard for any future resolving steps. I have no further concerns, LGTM.

terassyi (approved):
LGTM

PR #2110, opened 2026-09-11
pomerium: install UI dependencies with --ignore-scripts

shunki-fujita:
## What

- `pomerium/Dockerfile`: set `ENV NPM_CONFIG_IGNORE_SCRIPTS=true` in the build stage so that both the explicit `npm ci` and the one run by `make build-ui` (via upstream's `npm-install` target) skip dependency install scripts. (The first revision used `npm ci --ignore-scripts`, which did not cover the second `npm ci`; see the review thread.)
- `pomerium/TAG`: 0.33.1.1 -> 0.33.1.2 so CI builds the image on this PR.

## Why

The company-wide npm / pnpm security guideline recommends `ignore-scripts` so that a compromised dependency cannot run arbitrary code at install time. Among the images in this repository that build JavaScript, pomerium was the only one still running dependency lifecycle scripts (hubble-ui uses `--ignore-scripts`, argocd relies on the pnpm 10 default, grafana on upstream's `enableScripts: false`).

## Why the build keeps working

In the upstream `ui/package-lock.json` (v0.33.1), exactly one of 370 packages has an install script: `esbuild` 0.28.1. Its postinstall only hard-links the platform binary as a startup optimization and explicitly accounts for being skipped under `--ignore-scripts`. The platform packages (`@esbuild/linux-x64`, `@esbuild/linux-arm64`) are optional dependencies pinned in the lockfile and are installed by `npm ci`, so `make build-ui` (`ts-node ./scripts/esbuild.ts`) runs unchanged. The root `package.json` has no `preinstall` / `postinstall` / `prepare` scripts.

## Verification

- CI builds the image on this PR (TAG bumped). The `build_container` job for pomerium succeeds, and its log no longer shows the npm `allow-scripts` warning for `esbuild@0.28.1 (postinstall: node install.js)` on either `npm ci`.
- npm is the one bundled with Node 24.19.0 (npm 11.17.0), above the guideline's minimum of 11.11.0.


pomerium/Dockerfile
```diff
@@ -35,7 +35,8 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 COPY --from=source /pomerium.tar.gz /tmp/pomerium.tar.gz
 RUN tar zxf /tmp/pomerium.tar.gz --strip-components 1 && rm /tmp/pomerium.tar.gz
 
-RUN cd ui; npm ci
+# esbuild's postinstall is only an optional optimization, so skipping scripts is safe.
+RUN cd ui; npm ci --ignore-scripts
 RUN make build-ui
```
ystkfujii, line 41: [Q] Is this okay?

https://github.com/pomerium/pomerium/blob/01af7466fa3b26b5e5369df49f367737d81ad18e/Makefile#L197

ystkfujii: It looks like the install script is still being executed 👀 

https://github.com/cybozu/neco-containers/actions/runs/34576759143/job/103191119553#step:5:366

rkwmt (approved):

PR #2114, opened 2026-09-16
cilium: v1.17.12.4

hanapedia:
Bump cybozu version to include the following patch:
https://github.com/cybozu-go/cilium/pull/11


cilium-certgen/TAG
```diff
@@ -1 +1 @@
-0.4.5.1
+0.4.4.1
```
terassyi: cilium-certgen コンテナを新規にビルドする必要あるんですか？

cilium/e2e/Makefile
```diff
@@ -1,12 +1,17 @@
...
+CILIUM := $(BIN_DIR)/cilium
 CILIUM_DIR := ../src/cilium
 
 CILIUM_AGENT_IMAGE_TAG ?= ghcr.io/cybozu/cilium:$(shell cat ../TAG)
 HUBBLE_RELAY_IMAGE_TAG ?= ghcr.io/cybozu/hubble-relay:$(shell cat ../../hubble-relay/TAG)
+CILIUM_CLI_VERSION := 0.19.2
 
 .PHONY: setup
-setup:
-	aqua install
```
terassyi, line 9: Why did you remove `aqua install`?
I think it's because we changed to install Cilium CLI via curl in this PR.

Are there any reason to change this?

hanapedia: I just restored the files back to the point where it last built 1.17. To be more precise I just copied files from this branch: https://github.com/cybozu/neco-containers/tree/cilium-1.17.

I thought it was best to keep the switching between building 1.17 and 1.18 simple. The current flow is just copying files over branches.

terassyi: Ok, I got it.

terassyi (approved):
LGTM

PR #2115, opened 2026-09-17
cilium: v1.17.12.5

hanapedia:
- use latest commit on v1.17.12-cybozu branch.
- add simple check to validate that CILIUM_TARGET and the latest commit hash on -cybozu branch match.
- update docs


yokaze (approved):
LGTM

